PR #128, opened 2017-01-13
More travis tests - and a fix for classic Makefile QNormaliz "make install"

mkoeppe:
This pull request adds more travis tests (in particular some limited tests that what is installed by "make install" works) and catches a typo in the classic Makefile that was exposed by these tests.

By the way, all Travis builds involving "make distcheck" currently fail on branch "towards320" because the repository does not have the following files listed in examples/Makefile.am:

+       rationalES.intOut                       \
+       rationalLC.intOut                       \
+       rationalInt.intOut                      \

### DIFF
--- a/.travis-build.sh
+++ b/.travis-build.sh
@@ -43,51 +43,9 @@ case $BUILDSYSTEM in
 	fi
 	;;
 esac
-# Build Normaliz.
-cd $NMZDIR
+# Set up CoCoA if necessary for this build.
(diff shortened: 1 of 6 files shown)


w-bruns:
I had seen this. The problem was caused by git since "*.intOut" is in gitignore. I have now taken care that these files are tracked. They are contained in example since they occur explicitly in the NmzIntegrate manual.

PR #141, opened 2017-05-03
Fix in new compute method

sebasguts:


sebasguts:
I fixed a typo in the string translation of cone properties.

PR #147, opened 2017-06-25
Change nmz_interrupted to `volatile sig_atomic_t`

fingolfin:
Without the volatile, optimized code may not reload nmz_interrupted
from memory, leading to delays or failures to act upon a change in the
nmz_interrupted value.

The change from `int` to `sig_atomic_t` may not be strictly necessary
here, as the value in question is always in 0 and 1, and there should
be no race in writing it; but there seems no reason to not do it
"right" anyway.

See also:
* <https://www.securecoding.cert.org/confluence/display/c/SIG31-C.+Do+not+access+shared+objects+in+signal+handlers>
* <http://pubs.opengroup.org/onlinepubs/009695399/basedefs/signal.h.html>
* <http://en.cppreference.com/w/c/program/sig_atomic_t>

### DIFF
--- a/doc/Normaliz.tex
+++ b/doc/Normaliz.tex
@@ -6170,11 +6170,11 @@ The \verb|InterruptException| is discussed in the next section.
 
 In order to find out if the user wants to interrupt the program, the functions in \verb|libnormaliz| test the value of the global variable
 \begin{Verbatim}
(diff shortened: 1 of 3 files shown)


w-bruns:
Thanks! Done.

Winfried

PR #165, opened 2017-11-27
Added flint to autotools build

sebasguts:


sebasguts:
@mkoeppe could you have a look please? I am not an expert in autotools, so it would be nice to have a second opinion.

PR #182, opened 2018-04-03
Unified getter

sebasguts:
Unified getter methods for cone properties.


sebasguts:
This would also include the changes provided in #150.

PR #192, opened 2018-05-12
New install

sebasguts:
* Automatically download Py(Q)Normaliz for distribution
* Remove python dirs from repo.


mkoeppe:
Instead of the commit sha, could of course create tags corresponding to version numbers in the PyNormaliz and PyQNormaliz repositories and refer to those.

PR #199, opened 2018-06-07
Refactored flint-checking in configure

### LINKED ISSUE #198: Compiling normaliz 3.6.0 with flint2 installed leads to error
On a machine with flint 2.4.5 installed, compiling normaliz 3.6.0 runs into a compile error:
```
$ ./configur && make
...
In file included from libnormaliz/enumeration.cpp:28:
./libnormaliz/HilbertSeries.cpp:226:9: error: use of undeclared identifier 'fmpz_poly_cyclotomic'
        fmpz_poly_cyclotomic(cyc, (ulong) n);
        ^
1 error generated.
```

I am told this is because my version of flint is too old. That's fine -- but I think that the configure script's check for flint then should either check the version of flint, and/or check for the availability of that symbol, at least when it is auto-detecting flint (i.e. the user did not give an explicit `--with-flint`); and it then should *not* use flint if the found flint version is not suitable.

Also, while I am at it: I think it'd be helpful if the configure script printed something like "checking for flint ... yes / no", which IMHO is a best practice for configure scripts.

sebasguts:
Closes #198.

### DIFF
--- a/.travis-build.sh
+++ b/.travis-build.sh
@@ -118,7 +118,7 @@ case $BUILDSYSTEM in
     *)
 	# autotools, no Flint
 	./bootstrap.sh || exit 1
-	./configure $CONFIGURE_FLAGS --prefix="$INSTALLDIR" --disable-flint || ( echo '#### Contents of config.log: ####'; cat config.log; exit 1)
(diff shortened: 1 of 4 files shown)


sebasguts:
Please wait until all checks have passed before merging this.

PR #237, opened 2019-07-21
Make cone.h, fullcone.h compatible with older C++ versions

fingolfin:
With these changes, I can use normaliz again with older C++ compilers.

### DIFF
--- a/source/libnormaliz/cone.h
+++ b/source/libnormaliz/cone.h
@@ -82,7 +82,7 @@ struct CONVEXHULLDATA{
     size_t nrTotalComparisons; // counts the comparisons in the current computation
         
 
-    list<FACETDATA<Integer>> Facets;  // contains the data for Fourier-Motzkin and extension of triangulation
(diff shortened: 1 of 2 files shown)


w-bruns:
O.k., no problem. The >> were more or less accidental. Usually we avoid these.

By the way, we have have also got suggestions to go to C++17. Normaliz itself needs only C++11, but optional packages may need a newer version.

fingolfin:
That would be really sad and annoying, as there are many reasons why people may not be able to use a C++17 compatible compiler. I also don't follow the reasoning: If optional packages want to use C++17, why should that prevent people who don't use these optional packages from using Normaliz with an older C++ compiler?

Personally, I'd also recommend the authors of those optional packages to reconsider this and perhaps restrict to C++14 or even C++11; while I understand that using C++17 can be convenient for the code author, it is a major stumbling block for a large group of users, and I seriously wonder if the gained convenience in writing the code outweighs the benefits? Of course I don't know the details, but it would be the first time I see where this was the case. Of course for everything there is a first time... :-)

PR #242, opened 2019-07-23
Add ClassGroup to getVectorConeProperty

fingolfin:
Resolves #280


w-bruns:
I would not like to do this. It is true, the return type of the function getClassGroup() is a vector, but the computed vector is only an encoding of a complex structure. We return it as a vector with entries of type Integer only for simplicity. It could be avector<int> or vector<mpz_class> if one wants to be on the safe side, and. The other vectors belonging to VectorConeProperty are true vectors that stand for themselves.

fingolfin:
That is also fine by me but then this property should not be marked as a vector property. As it is, I have to add a workaround to NormalizInteface to deal with this mismatch.

fingolfin:
(And I'll keep treating it as a vector there anyway, as that is all that matters for the kernel part. The high level code using the kernel wrappers can then of course convert it to something else, e.g. a GAP group object).

PR #251, opened 2019-08-12
Fix more compiler warnings, and perhaps some bugs

fingolfin:
There are a few places in here where this PR changes semantics, and/or where the changes might have uncovered issues; I'll flag them so that they can be reviewed more closely.

### DIFF
--- a/source/libnormaliz/cone.cpp
+++ b/source/libnormaliz/cone.cpp
@@ -6911,8 +6911,6 @@ void Cone<Integer>::make_face_lattice(const ConeProperties& ToCompute){
     }
     if(verbose)
         verboseOutput() <<"Cosimplicial gens " << nr_simpl << " of " << nr_gens << endl;
-    bool ise_simple_vert=(10*nr_simpl>nr_gens);
-    
     
     vector<size_t> prel_f_vector(dim+1,0);
     
--- a/source/libnormaliz/cone.cpp
+++ b/source/libnormaliz/cone.cpp
@@ -3505,9 +3505,10 @@ ConeProperties Cone<Integer>::compute(ConeProperties ToCompute) {
     ToCompute.check_sanity(inhomogeneous);
     if(inhomogeneous){
         if(Grading.size()==0){
-            if(ToCompute.test(ConeProperty::DefaultMode))
+            if(ToCompute.test(ConeProperty::DefaultMode)) {
                 ToCompute.reset(ConeProperty::HilbertSeries);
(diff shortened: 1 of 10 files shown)


source/libnormaliz/cone.cpp
```diff
@@ -906,6 +907,8 @@ void Cone<Integer>::process_multi_input_inner(map< InputType, vector< vector<Int
                 break;
             case Type::polyhedron:
                 inhom_input=true;
+                nr_cone_gen++;
+                break;
```
fingolfin: This preserves semantics by making explicit what happens; but please verify that this is actually the intended behavior.

source/libnormaliz/cone.cpp
```diff
@@ -916,13 +919,16 @@ void Cone<Integer>::process_multi_input_inner(map< InputType, vector< vector<Int
             case Type::normalization:
             case Type::cone_and_lattice:
                 nr_cone_gen++;
+                nr_latt_gen++;
+                break;
```
fingolfin: This preserves semantics by making explicit what happens; but please verify that this is actually the intended behavior.

source/libnormaliz/cone.cpp
```diff
@@ -1550,6 +1556,8 @@ void Cone<Integer>::prepare_input_generators(map< InputType, vector< vector<Inte
                 LatticeGenerators.append(it->second);
                 if(BasisMaxSubspace.nr_of_rows()>0)
                     LatticeGenerators.append(BasisMaxSubspace);
+                Generators.append(it->second);
+                break;
```
fingolfin: This preserves semantics by making explicit what happens; but please verify that this is actually the intended behavior.

source/libnormaliz/cone.cpp
```diff
@@ -3496,9 +3505,10 @@ ConeProperties Cone<Integer>::compute(ConeProperties ToCompute) {
     ToCompute.check_sanity(inhomogeneous);
     if(inhomogeneous){
         if(Grading.size()==0){
-            if(ToCompute.test(ConeProperty::DefaultMode))
+            if(ToCompute.test(ConeProperty::DefaultMode)) {
                 ToCompute.reset(ConeProperty::HilbertSeries);
                 ToCompute.reset(ConeProperty::NoGradingDenom);            
+            }
```
fingolfin: This changes semantics, based on the assumption that the indentation showed the actual intent of the code. Please verify that this guess is right.

PR #252, opened 2019-08-12
Update .gitignore, turn on more compiler warnings, fix a bug in source/Makefile.am

fingolfin:
This adds `-Wall -pedantic` to the compiler flags, which is how I founds most of the warnings fixed by a previous PR. Actually, I used a few other flags for that as well, but they are different between clang and gcc, so for now, let's enable at least this common minimum set.


fingolfin:
I just made a minor update to this PR. If somebody can cancel this redundant Travis job, please do: https://travis-ci.org/Normaliz/Normaliz/builds/570856491

PR #254, opened 2019-08-12
Replace explicit ::iterator with auto keyword

fingolfin:
... in most places where it is possible. The resulting code is often
easier to read and maintain. This also enables future refactoring,
e.g. use of ranged for-loops.

Caveat: use of `auto` may not work with OpenMP, but I can't test that locally. So the Travis tests using OpenMP may very well fail, in which case I can update this PR. But before I invest that work, it'd be good to know if there is any interest in this at all, otherwise I'll just drop it.

### DIFF
--- a/source/input.cpp
+++ b/source/input.cpp
@@ -673,7 +673,6 @@ map <Type::InputType, vector< vector<Number> > > readNormalizInput (istream& in,
     bool we_have_a_polynomial=false;
 
     map<Type::InputType, vector< vector<Number> > > input_map;
-    typename map<Type::InputType, vector< vector<Number> > >::iterator it;
 
     in >> std::ws;  // eat up any leading white spaces
     int c = in.peek();
--- a/source/libnormaliz/HilbertSeries.cpp
+++ b/source/libnormaliz/HilbertSeries.cpp
@@ -261,9 +261,8 @@ long lcm_of_keys(const map<long, denom_t>& m){
 void HilbertSeries::compute_hsop_num() const{
         // get the denominator as a polynomial by mutliplying the (1-t^i) terms
         vector<mpz_class> hsop_denom_poly=vector<mpz_class>(1,1);
-        map<long,denom_t>::iterator it;
         long factor;
-        for (it=hsop_denom.begin();it!=hsop_denom.end();++it){
+        for (auto it=hsop_denom.begin();it!=hsop_denom.end();++it){
             factor = it->first;
             denom_t& denom_i = it->second;
             poly_mult_to(hsop_denom_poly,factor,denom_i);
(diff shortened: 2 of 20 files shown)


w-bruns:
Syntactically o.k., but it fails at run time. My impression: at least one loop does not terminate properly. For example we see bad_alloc, usually an indication that memory is exhausted.

Is it possible to start a new branch with the present master and this pull request (in order not to change the master branch) ? Otherwise I merge, pull, start a new branch, reset master and push again.

w-bruns:
Before I forget: thanks for cleaning this up!

A remark on history: in the first years of C++ coding there was no auto. So the older parts of the code have those complicated iterator definitions.

w-bruns:
With the second commit some builds were successful, the three Linux ones without OpenMP and one more. Others failed because of syntax problems. I guess the latter were cured with the last commit.

fingolfin:
I am glad you like the general direction of this PR! That means it is worthwhile for me to debug and fix it -- I was aware of the crash, but just didn't want to bother with it before I knew whether you liked the general idea of this (if you didn't, then it would have been a waste of time to debug it).

Anyway, of course you can also try it out. The easiest way would be to simply clone my form of the Normaliz repository, and then check out my branch `mh/use-auto`. Slightly more advanced is to work within an existing clone of the main Normaliz repository, but add my repository as a `git remote`, then access it. Something like this should do that, executed inside an existing Normaliz.git clone:
```
git remote add fingolfin https://github.com/fingolfin/Normaliz
git fetch fingolfin # ask git to fetch all my branches
git checkout mh/use-auto
```

w-bruns:
There is a problem in full_cone.cpp in find_new_facets(), which is a very old part of the program (but works very well). With the following patch all tests go through:

```
        if(potential_ranktest)
            convert(Generators_float,Generators);
    }

    #pragma omp parallel private(jj) // <----- we need this private (or must take care of it in another way)
    {
    size_t i,j,k,nr_zero_i;
    boost::dynamic_bitset<> subfacet(dim-2);
    jj = Neg_Subfacet_Multi_United.begin();  // <---- we need this initialization
    size_t jjpos=0;
    int tn = omp_get_ancestor_thread_num(omp_start_level+1);
```
Please add another commit to your pull request.

w-bruns:
Nicer alternative:
```
    #pragma omp parallel
    {
    size_t i,j,k,nr_zero_i;
    boost::dynamic_bitset<> subfacet(dim-2);
    auto jj = Neg_Subfacet_Multi_United.begin();
    size_t jjpos=0;
```

fingolfin:
@w-bruns OK, thanks! PR update, let's see.

PR #256, opened 2019-08-14
Use ranged for loops from C++11

fingolfin:
This PR changes the code to use C++11 ranged for loops. This makes a lot of the code easier to read (IMHO at least). The first commit was generated using the `clang-tidy` tool, but then a lot of other loops were changed with manual work.

AFAIK, ranged for loops are only supported in OpenMP 5.0, so I'd not be surprised if I broke compilation when using OpenMP; Travis will tell us, and then affected loops can be changed back to "old style".


w-bruns:
A real improvement! Thanks. 
Though it is not tested on Travis, I am pretty confident that the C++ compiler that I use for MS Windiws will get along with the changes.

PR #260, opened 2019-08-21
Simplify configure.ac

saraedum:
* Renamed files to follow GNU guidelines
* Simplified configure.ac (I don't think we should try to be so smart here. Just rely on people setting CXXFLAGS, LDFLAGS, … correctly if things happen to live in non-standard places.)


saraedum:
This is mostly untested but is what I think a configure.ac should essentially look like. What do you think @w-bruns @isuruf @videlec?

videlec:
The travis script seems broken. The reason is that it runs the command
```
./configure $CONFIGURE_FLAGS  --prefix=${INSTALLDIR} --with-cocoalib=${INSTALLDIR} --with-nauty=${INSTALLDIR} --with-flint=${INSTALLDIR} --disable-shared
```
independently of what is actually available.

saraedum:
There seems to be a misunderstanding about the purpose of version info. Let me also addresses that

configure.ac
```diff
@@ -1,313 +1,127 @@
...
 
 AC_PREREQ(2.59)
+
 AC_INIT(Normaliz, 3.7.4)
+
 AC_SUBST(LIBNORMALIZ_VERSION_MAJOR, 3)
 AC_SUBST(LIBNORMALIZ_VERSION_MINOR, 7)
 AC_SUBST(LIBNORMALIZ_VERSION_PATCH, 4)
 AC_SUBST(LIBNORMALIZ_VERSION_STRING, "$PACKAGE_VERSION")
-AM_INIT_AUTOMAKE(foreign)
```
fingolfin: I really would suggest to keep the `foreign` here. Slavishly following the GNU conventions seems pointless. Of course it doesn't hurt to e.g. rename the change log file, but adding symlinks `README` and `NEWS` is simply silly (no offense intended).

INSTALL
```diff
@@ -1,288 +1,368 @@
...
-   (http://scip.zib.de/academic.txt).
-   
-* Unpack it and then compile it with
- 
-   make ZLIB=false GMP=false READLINE=false scipoptlib
-   
-   (To compiling scip for 32bit systems, add also ARCH=x86 to the make.)
-   
-   If your C++ compiler is too old, you may need to add
-   CXXFLAGS="-std=c++0x -fPIC"
```
fingolfin: So you have removed all these instructions, by replacing INSTALL with the stock GNU INSTALL file. That seems like a bad idea to me.

configure.ac
```diff
@@ -1,313 +1,127 @@
...
-AC_ARG_ENABLE([openmp],
-    [AS_HELP_STRING([--enable-openmp@<:@=ARG@:>@],
-      [enable parallelization with OpenMP @<:@default=check@:>@])],
-    [:],
-    [enable_openmp=check])
-AS_IF([test "x$enable_openmp" != xno],
-  [AX_OPENMP([have_openmp=yes],
-	     [have_openmp=no])])
+# We use GMP's C++ wrapper but we also interface with GMP's C interface directly.
+AC_CHECK_LIB([gmp], [__gmpz_init], [have_gmp=yes], AC_MSG_ERROR([GMP library not found]))
```
fingolfin: So `--with-gmp` has been removed without a replacement? Is that intentional? Can you explain the motivation?

Also, you then need to adjust places using it, i.e. at least `install_normaliz.sh`

configure.ac
```diff
@@ -1,313 +1,127 @@
...
-  [enable_openmp=yes],
-  [AS_IF([test "x$enable_openmp" = xyes],
-    [AC_MSG_ERROR([openmp not available])],
-    [AS_IF([test "x$enable_openmp" != xno],
-        [AC_MSG_NOTICE([Parallelization with OpenMP will not be available because no suitable compiler was found])])
-     enable_openmp=no])])
+# GMPXX does not contain anything that we can check for with AX_CXX_CHECK_LIB
+# so we just check for something from the standard library, i.e., that -lgmpxx
+# goes through
+AX_CXX_CHECK_LIB([gmpxx], [std::runtime_error::what () const], [have_gmpxx=yes], AC_MSG_ERROR([GMP library for C++ not found]), [-lgmp])
```
fingolfin: Could you elaborate on that? Why can't you check for `mpz_class`?

configure.ac
```diff
@@ -1,313 +1,127 @@
...
-	    break
-	   ],
-	   [have_scip=no])
-     done
-     LIBS="$LIBS_SAVED"
-     if test $have_scip != yes ; then
-	AC_MSG_RESULT($have_scip)
-     fi
+# We can use SCIP (Solving Constraint Integer Programs) for cone decomposition
+AC_ARG_WITH([scip], AS_HELP_STRING([--with-scip=FLAGS], [Use SCIP for cone decomposition; the FLAGS specify the linker flags necessary, e.g., -lscip. Typically, a subset of -lscip -lz -lreadline -lscipopt -lnlpi.cppad -llpispx -lsoplex or similar is going to make this work.]))
```
fingolfin: So you both removed the convenience code which made it a bit easier to link against SCIP, *and* also the INSTALL instructions related to SCIP. That strikes me as potentially problematic, though I never tried to use SCIP with Normaliz myself, so perhaps there is a reason why all this code and the instructions should go -- but then IMHO the pull request should explicitly explain them.

I also wonder: Did you test the whole thing with SCIP at all?

configure.ac
```diff
@@ -1,313 +1,127 @@
...
-AC_SUBST(NAITY_CPPFLAGS)
-AM_CONDITIONAL(NMZ_NAUTY, test x$enable_nmznauty = xyes)
-AS_IF([test "x$have_nauty" = xyes],
-  [AC_DEFINE(NMZ_NAUTY)
-   DEFINE_NMZNAUTY="#define NMZ_NAUTY"
-   AC_SUBST([DEFINE_NMZNAUTY])])
+# We can optionally build with nauty.
+AC_ARG_WITH([nauty], AS_HELP_STRING([--with-nauty], [Use nauty for nmzInautomorph]))
+AS_IF([test "x$with_nauty" = "x"], [with_nauty=no])
+AS_IF([test "x$with_nauty" != "xno"], [
```
fingolfin: But you never use the value of `with_nauty`. If you don't want to support specifying custom installation paths for dependencies anymore (but why?), then shouldn't this just use `AC_ARG_ENABLE` instead? (Same of course applies to virtually all the other dependencies).

w-bruns:
I don't know autotools well enough to make a useful comment. The rewriting of configure.ac started from Julian's observation that both "with/Without" and "enable/disable" were used in the file essentially written by Matthias Köppe. No doubt that this duplication should be avoided.

As far as Scip is concerned: we can remove it. Two years ago I found (as a byproduct of another algorithm) a method that replaces Scip in Normaliz, is faster and gives better results. I kept the option for political reasons and also to have a foot in the door in case it could become useful again. The latter is rather unlikely now, and in any case we can resort to opening an old version that has it.

configure.ac
saraedum: I like to follow this standard. If packagers find all the files that are usually around it makes their life easier as they do not have to think about how stuff might be named.

INSTALL
saraedum: That was not intentional.

configure.ac
saraedum: I removed it because I don't see why it should be used to specify a search path. That's what LDFLAGS is good for. As for `install_normaliz.sh`, sure that needs to be adapted, thanks for pointing that out. (Feel free to patch it up.) I don't really understand what that script is good for though.

configure.ac
saraedum: `mpz_class` is not a symbol, it's a typedef that resolves to something very complicated.

configure.ac
saraedum: No. As I wrote, I did not test this. At the moment, this was just about pointing out how a `configure.ac` should roughly look like.

configure.ac
saraedum: The (probably unwritten) standard says that `--enable` is for enabling/disabling features that are usually independent of dependencies. `--with` is for building against dependencies. The value of `--with` is usually not used. You just look for `with/without` and ignore the actual value typically.

saraedum:
To clarify, it's fine to use `--with` and  `--enable` in parallel but usually it's unnecessary. What really surprised us for a bit when packaging for conda-forge was that `--without-something` would result in `with_something=no` (that's just how autoconf works) which in turn would add `-Lno/lib` to the search path.

saraedum:
@w-bruns if we were to drop scip support could we use a standard INSTALL file?

INSTALL
```diff
@@ -1,288 +1,368 @@
...
-MPFR,  http://www.mpfr.org/mpfr, version 4.0.1
-Flint, http://www.flintlib.org/flint, version 2.5.2
-nauty, http://users.cecs.anu.edu.au/~bdm/nauty, version 27rc2
+     ./configure CC=/usr/local2/bin/gcc
 
-For algebraic polyhedra additionally:
+causes the specified 'gcc' to be used as the C compiler (unless it is
+overridden in the site shell script).
 
-MPFR and Flint as above
```
saraedum: I think a check for mpfr is still missing from `configure.ac`.

INSTALL
```diff
@@ -1,288 +1,368 @@
...
-arb, https://github.com/fredrik-johansson/arb/archive/, version 2.16.0
```
saraedum: Is arb used directly or only indirectly through e-antic? @w-bruns

INSTALL
```diff
@@ -1,288 +1,368 @@
...
 
-They installation can in addition be controlled by several environment variables:
+'--help=short'
+'--help=recursive'
+     Print a summary of the options unique to this package's
+     'configure', and exit.  The 'short' variant lists options used only
+     in the top level, while the 'recursive' variant lists options also
+     present in any nested packages.
 
-NMZ_COMPILER: overrides the standard choice of compiler (g++ on Linux, clang++ on Mac)
```
saraedum: I believe that all of these variables should be handled by the standard switches that `./configure` provides already.

INSTALL
```diff
@@ -1,288 +1,368 @@
...
-within the Normaliz directory.
-
-1) Go to the source directory
-
-2) Check the configuration in Makefile.configuration, especially the locations of
-the optional packages and their versions.
-
-3) Say 
-
-        make -f Makefile.classic
```
saraedum: What is the benefit of this Makefile over the automake Makefile? @w-bruns

INSTALL
```diff
@@ -1,288 +1,368 @@
...
-        make -f Makefile.classic Flint=no   # deactivates Flint
-        make -f Makefile.classic EANTIC=no  # deactivates e-antic
-        make -f Makefile.classic NMZ_DEVELOP=no  # deactivates development version
-
-You must make sure that everything is compiled with the same set of options!
-In doubt use "make -f Makefile.classic clean" beforehand.
-
-Similarly
-
-        make -f Makefile.classic maxsimplex
```
saraedum: Is this different from `cd source; make maxsimplex/maxsimplex` with autotools?

configure.ac
fingolfin: You are right, the GNU autoconf manual explicitly writes (so it is written :-)) that `--with` is for packages/dependencies, and `--enable` for feature. OK, fine.

But that the value of `--with` is usually not used is simply not true, based on dealing with this stuff for the past twenty years. For example, let's look at the GCC configure -- arguably, GCC is maybe *the* GNU project. It does this:
```
...

Optional Packages:
  --with-PACKAGE[=ARG]    use PACKAGE [ARG=yes]
  --without-PACKAGE       do not use PACKAGE (same as --with-PACKAGE=no)
  --with-build-libsubdir=DIR  Directory where to find libraries for build system
  --with-system-zlib      use installed libz
  --with-mpc=PATH         specify prefix directory for installed MPC package.
                          Equivalent to --with-mpc-include=PATH/include plus
                          --with-mpc-lib=PATH/lib
  --with-mpc-include=PATH specify directory for installed MPC include files
  --with-mpc-lib=PATH     specify directory for the installed MPC library
  --with-mpfr-dir=PATH    this option has been REMOVED
  --with-mpfr=PATH        specify prefix directory for installed MPFR package.
                          Equivalent to --with-mpfr-include=PATH/include plus
                          --with-mpfr-lib=PATH/lib
  --with-mpfr-include=PATH
                          specify directory for installed MPFR include files
  --with-mpfr-lib=PATH    specify directory for the installed MPFR library
  --with-gmp-dir=PATH     this option has been REMOVED
  --with-gmp=PATH         specify prefix directory for the installed GMP
                          package. Equivalent to
                          --with-gmp-include=PATH/include plus
                          --with-gmp-lib=PATH/lib
  --with-gmp-include=PATH specify directory for installed GMP include files
  --with-gmp-lib=PATH     specify directory for the installed GMP library
  --with-stage1-libs=LIBS libraries for stage1
  --with-stage1-ldflags=FLAGS
                          linker flags for stage1
  --with-boot-libs=LIBS   libraries for stage2 and later
  --with-boot-ldflags=FLAGS
                          linker flags for stage2 and later
  --with-isl=PATH         Specify prefix directory for the installed isl
                          package. Equivalent to
                          --with-isl-include=PATH/include plus
                          --with-isl-lib=PATH/lib
  --with-isl-include=PATH Specify directory for installed isl include files
  --with-isl-lib=PATH     Specify the directory for the installed isl library
...
```

Of course one can still say that one does not want to support a prefix argument. And if this `configure.ac` file was being created new and from scratch, I'd be fine with that. But you are not proposing to remove functionality, without a clear argument for why, i.e., what is gained by this. At least it didn't become clear to me, but perhaps you can elaborate?

configure.ac
fingolfin: OK. But as @w-bruns wrote that SCIP support perhaps can be removed completely (did I get that right), that may be the better option -- i.e., let's first remove SCIP support completely in a separate PR, that then also makes reviewing easier :-). I can look into doing that (but would be best if @w-bruns first confirms that I did not misunderstand :-).

configure.ac
fingolfin: I've worked as a package for 20 years, and you always have to check for which of these files exist and in which variant, and it is one of the easiest aspects of the whole process, so I really don't agree with this justification.

configure.ac
saraedum: GCC uses these paths because GCC is very complex (different parts of GCC build against varying dependencies I guess.) Most projects do not use this afaik.

The problem is that `--with-gmp=path` is an alias for `LDFLAGS=-Lpath/lib` `CPPFLAGS=`-Ipath/include`. I think that people should just set their LDFLAGS  and CPPFLAGS correctly; that's certainly how distributions build packages. We can keep the current behaviour but it creates a fair bit of complexity in the configure.ac (you need to handle no and yes differently and patch LDFLAGS and CPPFLAGS.) Since we're breaking the interface anyway by removing the `--enable` flags, we can also change this I believe.

Note that `--with-flint=$PREFIX` (that's what existing build scripts for distributions will do) still works. The argument is ignored but their LDFLAGS and CPPFLAGS certainly include the correct paths already anyway.

configure.ac
fingolfin: Relevant should be a symbol from libgmpxx which is needed by Normaliz. So I tried linking without it, and the only missing symbol it complained about was `operator>>(std::__1::basic_istream<char, std::__1::char_traits<char> >&, __mpq_struct*)`. So that's what I'd test for. If `AX_CXX_CHECK_LIB` can't do that, I wouldn't use `AX_CXX_CHECK_LIB`, and just roll the code for it myself (something like `mpq_class tmp; std::cin >> tmp;` should probably do the trick).

configure.ac
saraedum: I have no strong opinion on this. I just felt that this would streamline things a bit.

configure.ac
fingolfin: Fair enough. So, if everybody is happy with dropping all the optional prefixes from the `--with` arguments (and adapting code relying on that, and enhance the build instructions to explain how to do it instead, etc.), I personally don't mind -- I have no troubles building Normaliz either way. But in my experience (i.e., maintenance of other software using configure scripts, with similar options, and the support feedback we get from users) , a lot of people expect that they can do e.g. `--with-gmp=PATH`, and if this works, they are confused, and ask us for help (and it doesn't matter if we explain how to do it somewhere --- few people read those instructions, sadly. Well, I usually don't anymore, too ;-).

So, it's a tradeoff: simplicity versus user-friendliness. Is the extra simplicity worth the fact that it may be harder for some people to compile Normaliz? (Though of course this PR may simplify other things, so the overall balance probably needs to be taken into account, too)

configure.ac
saraedum: I think that's complicating things too much as we're not searching for all the used symbols elsewhere. I don't think you'll find a system that has a working `libgmp` but a `libgmpxx` that is missing that symbols. But feel free to add such a check if you think it's a good idea.

INSTALL
w-bruns: I would like to keep both possibilities, via autotools and via Makefile.classic

INSTALL
w-bruns: I can't make a fully static Linux binary via autotools. For this purpose I need Makefile.classic.
But perhaps you know a way to produce fully static binaries via autotools.

INSTALL
w-bruns: It is. But perhaps it makes sense to list it here. I am not sure.

INSTALL
w-bruns: It is of course possible to keep Makefile.classic on the private side and not mention it in INSTALL.  Sp you can remove it from INSTALL as well as from the distribution via EXTRADIST. But I would like to keep it in the git repository.

w-bruns:
Let us drop SCIP if it hasn't been done yet. I can do it in the source code and in the manual.

INSTALL
saraedum: Sounds good. If it's useful to you we should keep it in the repo.

INSTALL
saraedum: This seems to work for me:
```
❯ LDFLAGS="--static" ./configure --disable-shared --enable-static
❯ make
❯ ldd ./source/normaliz
	not a dynamic executable
```

saraedum:
@w-bruns I can rebase this PR once you dropped SCIP support.

w-bruns:
Very good.
I will remove Scip from the source tomorrow morning.

w-bruns:
A Scip free version of master pushed to GitHub. (Scip only mentioned in INSTALL anymore, but that will change anyway.)

w-bruns:
Scip support has been dropped completely in master on GitHub.

fingolfin:
(BTW, despite all my nitpicking here, I like the direction of this PR -- thank you for working on it)

w-bruns:
I would like to include the new version of the scripts to version 3.8.0 that is planned for September 10. Therefore I would be grateful if this pull request could be finished the next days.

INSTALL
saraedum: So, is there now a need still for a custom INSTALL?

configure.ac
fingolfin: Well, I find it outright confusing to have these pointless symlinks there. There is a reason why the "foreign" option exists, and why many (most?) non-GNU projects use it.

source/Makefile.am
```diff
@@ -55,12 +44,12 @@ noinst_HEADERS += input.cpp options.cpp
...
 noinst_PROGRAMS = maxsimplex/maxsimplex
 maxsimplex_maxsimplex_SOURCES = maxsimplex/maxsimplex.cpp
-maxsimplex_maxsimplex_LDADD = libnormaliz.la
+maxsimplex_maxsimplex_LDADD = libnormaliz.la -lgmp -lgmpxx
 
 # Uninstalled binary dynamic
 noinst_PROGRAMS += dynamic/dynamic
 dynamic_dynamic_SOURCES = dynamic/dynamic.cpp
-dynamic_dynamic_LDADD = libnormaliz.la
+dynamic_dynamic_LDADD = libnormaliz.la -lgmp -lgmpxx
```
fingolfin: That doesn't seem right -- libnormaliz.la should encode all transitive requirements, precisely so people using it (via GNU libtool) don't need to manually add them.

fingolfin: This may in particular actively break code linking against libnormaliz. I.e., looks like a blocker to me; but perhaps there is a rationale, that I then would love to understand

w-bruns:
Yesterday I wrote that I would like to include the revised scripts in Normaliz 3.8.0. But there is really no urgent need to do this. We can take our time and test them in all environments, especially with PyNormaliz and in Sage. The old scripts work well, despite the weak points they may have.

w-bruns:
I had a brief look at then Travis logs. Unfortunately a green cross does not imply that a particular build is completely o.k.  For example, 823.6 builds Noramliz without CoCoALIb though this library is asked for in the build.Also the tests for nauty are not done.

source/Makefile.am
saraedum: If dynamic uses GMP's C++ wrapper through any of the code in the header files it pulls in, it should link against `libgmpxx`. Almost anything you do from `gmpxx.h` means that you use GMP which means that you should link against `libgmp`. That's my understanding how linking works here but I might be wrong and dynamic might actually not use any GMP at all. In that case we would be overlinking here.

saraedum:
@w-bruns yes, I noticed that travis broke badly. I'll need to have another look there.

w-bruns:
Let me explain where the problem is with a build like 823.6 is. As far as I can see, thze wanted packages are found, compiled and installed. Then Normaliz is compiled, but the packages have not been found by configure so that the preprocessor variables are NOT set accurately. The tests depend on them, and if a preprocessot variable is not set, then the corresponding tests are not done. So we can end up with a green checkmark, although the desired configuration was not built. the tests cannot see this.

Therefore I always checked the Travis log files by hand if this was necessary /not so  often). A way out could be to set the preprocessor variables according to the desired build before the tests are selected. I guess this is done by make check.

PR #263, opened 2019-08-27
Remove remaining references to SCIP and cmake

fingolfin:


w-bruns:
Scip appears in the name of a subdirectory of test because these tests originally depended on Scip. They do no longer, and therefore they should not be deleted. To avoid confusion, I will change the names.

PR #265, opened 2019-09-10
configure.ac: some mild cleanup and refactoring

fingolfin:
- require autoconf 2.68, released in 2010; older versions have many bugs
- move AC_CONFIG_FILES to the end, where it usually sits
- replace obsolete AC_PROG_LIBTOOL by LT_INIT
- specify explicit AC_CONFIG_AUX_DIR;  in the future, we could change it from . to cnf
- replace shell comments `#` by m4 comments `dnl` in a few places

This is a very moderate change. The next step will be to replace the various package checks by better ones (which e.g. properly support `--without-FOO`; and which also give an error if `--with-FOO` is explicitly given, but `FOO` does not work / is missing).

As such, this is the start of (= first PR of) an alternative to PR #260.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,29 +1,47 @@
-#                                               -*- Autoconf -*-
-# Process this file with autoconf to produce a configure script.
+dnl
+dnl Process this file with autoconf to produce a configure script.
+dnl
 
-AC_PREREQ(2.59)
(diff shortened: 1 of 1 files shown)


w-bruns:
Can also go into 3.8.1 ...

PR #268, opened 2019-09-13
Refactor boost::dynamic_bitset<> use

fingolfin:
The first commit in this PR adds a new header file `source/libnormaliz/dynamic_bitset.h` which currently essentially contains this:
```C++
#include <boost/dynamic_bitset.hpp>
namespace libnormaliz {
typedef boost::dynamic_bitset<> dynamic_bitset;
}
```
The first commit also changes all the existing code to include the new header, instead of including the boost header directly.

The second commit then changes all the source code to use "our" `dynamic_bitset` instead of `boost::dynamic_bitset<>` (of course for now they are the same).

The ultimate motivation for this is that I would like to make it possible to compile Normaliz without Boost; right now, the dependency on Boost is the biggest issue I face in installing Normaliz on various systems. The biggest part of a patch achieving this goal (in number of lines changed) is taken care of by this PR. Thus, if I (or somebody else) wants to provide a drop-in replacement for `boost::dynamic_bitset` (or rather, for the subset of the `boost::dynamic_bitset` features used by Normaliz), they now only need to edit the new header file, but don't need to touch any other files.

### DIFF
--- a/doc/Normaliz.tex
+++ b/doc/Normaliz.tex
@@ -8190,7 +8190,7 @@ The return value is to be interpreted as follows: The entry for index $0$ is the
 \subsubsection{Face lattice and f-vector}
 \begin{Verbatim}
 vector<size_t> Cone<Integer>::getFVector()
-const map<boost::dynamic_bitset<>,int>& Cone<Integer>::getFaceLattice()
+const map<dynamic_bitset,int>& Cone<Integer>::getFaceLattice()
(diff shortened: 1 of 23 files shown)


w-bruns:
The only build that fails is the one without any optional library. I am quite sure you see immediately why it fails.
+

fingolfin:
@w-bruns fixed now

PR #270, opened 2019-09-17
Custom dynamic_bitset implementation

### LINKED ISSUE #221: boost presence should be checked in configure
If boost is not present compilation fails... would be nicer if configuration failed...
```
libtool: compile:  g++ -std=gnu++11 -DPACKAGE_NAME=\"Normaliz\" -DPACKAGE_TARNAME=\"normaliz\" -DPACKAGE_VERSION=\"3.7.0\" "-DPACKAGE_STRING=\"Normaliz 3.7.0\"" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -DPACKAGE=\"normaliz\" -DVERSION=\"3.7.0\" -DHAVE_CXX11=1 -DSTDC_HEADERS=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_MEMORY_H=1 -DHAVE_STRINGS_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_STDINT_H=1 -DHAVE_UNISTD_H=1 -DHAVE_DLFCN_H=1 -DLT_OBJDIR=\".libs/\" -DNMZ_FLINT=1 -DENFNORMALIZ=1 -I. -I../source -fopenmp -g -O2 -MT libnormaliz/enumeration.lo -MD -MP -MF libnormaliz/.deps/enumeration.Tpo -c libnormaliz/enumeration.cpp  -fPIC -DPIC -o libnormaliz/.libs/enumeration.o
In file included from ./libnormaliz/HilbertSeries.cpp:35:0,
                 from libnormaliz/enumeration.cpp:28:
./libnormaliz/vector_operations.h:31:36: fatal error: boost/dynamic_bitset.hpp: No such file or directory
compilation terminated.
Makefile:632: recipe for target 'libnormaliz/enumeration.lo' failed
make[1]: *** [libnormaliz/enumeration.lo] Error 1
make[1]: Leaving directory '/home/travis/build/videlec/PyNormaliz/normaliz-3.7.0/source'
Makefile:422: recipe for target 'all-recursive' failed
make: *** [all-recursive] Error 1
The command "make" failed and exited with 2 during .
```

fingolfin:
This is an alternative implementation of dynamic_bitset, which would mean that one can compile Normaliz without having Boost installed. This is not quite done, though. Among the things to be done:

- [x] run benchmarks to see how this compares against `boost::dynamic_bitset`
- [ ] add some dedicated tests for this code
- [x] resolve some TODOs
- [x] decide whether this should replace `boost::dynamic_bitset`, or be a offered as an alternative
  - [ ] if we keep both, then `configure` should allow selecting which to use
  - [x] if we drop usage of the Boost version, then we can of course also remove all references to Boost
- [ ] perhaps tune the performance (e.g. one could use MMX/SSE/AVX/whatever, if available), but unless we are slower than `boost::dynamic_bitset`, I'd rather leave that for the future.

Above all: Get feedback from @w-bruns what he thinks about this, before I waste more time on this ;-).

Note that this does not implement all features of `boost::dynamic_bitset`, intentionally; I focused on implementing just what Normaliz uses (plus a few "obvious" extras for symmetry). On the one hand, that saves quite some work, on the other hand, code that is not used and is not tested tends to be wrong anyway ;-). I will happily implement more features if the need arises, though.

Since the dependency on Boost is removed, this resolves #221


w-bruns:
Excellent! Not too much can be missing since all Travis builds have been successful.

I am looking forward to the final version.

fingolfin:
@w-bruns well, I posted a list of issues... (and in the code, you can see a few TODO comments. One big question is this: should this replace boost::dynamic_bitset, or be a offered as an alternative? In either case, some additional work is needed.

Personally, I'd just ditch the use of Boost, but then, there might be reasons to keep it that I am not aware of.

As to performance testing: I had a look at the boost code, and I think we are pretty much doing the same, except that I added some optimizations to use compiler intrinsics for bit counting / scanning, which *might* make this implementation here a tad faster for some specific test cases. Anyway, if you are aware of some Normaliz test cases which are specifically sensitive to the performance of dynamic_bitset, please let me know, then I can test them. Otherwise, I'll have to resort to some artificial micro benchmarks, which probably won't be very insightful overall.

w-bruns:
I would not mind giving up boost::dynamic_bitset, but we should be sure that we do not lose efficiency. 
I will suggest some test cases. For the beginning  you could take A543.in, A541.in and lo6.in from the example directory and run them with -cis on the command line. (.i switches off any computation goal in the source file and -s activates "support hyperplanes".) I will put some more into playground and push them.

TODOs:

1) find_next: not used at present, but I think it would be good to have. I had thought about using it at one point.
2) limb_find_first: unclear to me what it means.
3) output operator <<: we should have it. It is good for debugging, even if it is not used in output at present . 

These are the TODOs I found.

w-bruns:
I have put m13.in, m14.in and m15.in into master/playground and pushed them to GitHub. These are computations of f-vectors. m15.in may take a while.

For them the 0-1-vectors are really the crucial objects.

fingolfin:
@w-bruns so what should I do -- run `normaliz -cis playground/m13` and so on, and time the result with each bitset implementation? Or perhaps I should use other options?

fingolfin:
So I used https://github.com/sharkdp/hyperfine to benchmark `source/normaliz -cis playground/m13` on a server not doing anything else. I also disabled OpenMP, to reduce fluctuations that happen when multiple threads are involved. There was no eAntic or flint or nauty, but I compiled it against cocoalib, because I had that sitting around anyway. Default compiler settings (`-g -O2`), assertions enabled. Might be interesting to re-run this with `-O3 -march=native -DNDEBUG`.

Here are the numbers I got; they pain a pretty clear picture, I think.

M13:
```
$ ARGS='-cis playground/m13' ; hyperfine --warmup 3 -m 20 "source/normaliz $ARGS" "../normaliz-dynamic_bitset/source/normaliz $ARGS"
Benchmark #1: source/normaliz -cis playground/m13
  Time (mean ± σ):     621.1 ms ±  50.0 ms    [User: 565.6 ms, System: 8.6 ms]
  Range (min … max):   554.7 ms … 743.3 ms    20 runs

Benchmark #2: ../normaliz-dynamic_bitset/source/normaliz -cis playground/m13
  Time (mean ± σ):     562.2 ms ±  41.5 ms    [User: 536.8 ms, System: 11.0 ms]
  Range (min … max):   496.5 ms … 606.6 ms    20 runs

Summary
  '../normaliz-dynamic_bitset/source/normaliz -cis playground/m13' ran
    1.10 ± 0.12 times faster than 'source/normaliz -cis playground/m13'
```

M14:
```
$ ARGS='-cis playground/m14' ; hyperfine --warmup 3 -m 20 "source/normaliz $ARGS" "../normaliz-dynamic_bitset/source/normaliz $ARGS"
Benchmark #1: source/normaliz -cis playground/m14
  Time (mean ± σ):      2.336 s ±  0.012 s    [User: 2.298 s, System: 0.025 s]
  Range (min … max):    2.315 s …  2.354 s    20 runs

Benchmark #2: ../normaliz-dynamic_bitset/source/normaliz -cis playground/m14
  Time (mean ± σ):      2.100 s ±  0.048 s    [User: 2.032 s, System: 0.022 s]
  Range (min … max):    2.012 s …  2.198 s    20 runs

Summary
  '../normaliz-dynamic_bitset/source/normaliz -cis playground/m14' ran
    1.11 ± 0.03 times faster than 'source/normaliz -cis playground/m14'
```

M15:
```
$ ARGS='-cis playground/m15' ; hyperfine --warmup 2 -m 10 "source/normaliz $ARGS" "../normaliz-dynamic_bitset/source/normaliz $ARGS"
Benchmark #1: source/normaliz -cis playground/m15
  Time (mean ± σ):     16.524 s ±  0.055 s    [User: 16.389 s, System: 0.112 s]
  Range (min … max):   16.451 s … 16.609 s    10 runs

Benchmark #2: ../normaliz-dynamic_bitset/source/normaliz -cis playground/m15
  Time (mean ± σ):     14.884 s ±  0.094 s    [User: 14.750 s, System: 0.110 s]
  Range (min … max):   14.689 s … 15.041 s    10 runs

Summary
  '../normaliz-dynamic_bitset/source/normaliz -cis playground/m15' ran
    1.11 ± 0.01 times faster than 'source/normaliz -cis playground/m15'
```

A543:
```
$ ARGS='-cis playground/A543' ; hyperfine --warmup 2 -m 10 "source/normaliz $ARGS" "../normaliz-dynamic_bitset/source/normaliz $ARGS"
TODO: will add once the numbers are in
```

lo6:
```
$ ARGS='-cis playground/lo6' ; hyperfine --warmup 2 -m 10 "source/normaliz $ARGS" "../normaliz-dynamic_bitset/source/normaliz $ARGS"
TODO: will add once the numbers are in
```


So the code in this PR is consistently 10% faster in this examples. That might because it uses compiler intrinsics for `count()` and `find_first()` (do either of those matter in those computations), or simply because the code is simpler than that in Boost, giving the compiler more optimization opportunities. Whatever it is, it seems we can switch to this new code w/o loosing performance; instead we might even gain some.

That said, other benchmarks could of course paint a different picture...

source/libnormaliz/dynamic_bitset.h
```diff
@@ -24,12 +24,449 @@
...
+    //
+
+    bool operator==(const dynamic_bitset &x) const {
+        return _total_bits == x._total_bits && _limbs == x._limbs;
+    }
+    bool operator!=(const dynamic_bitset &x) const {
+        return !(*this == x);
+    }
+    bool operator<(const dynamic_bitset &x) const {    // for use in std::set, std::map
+        // FIXME: the following is probably not compatible with boost::dyanmic_bitset; do we care?
```
fingolfin: This is the only TODO/FIXME comment left... I think we only need this method so that we can put dynamic_bitset instances into `std::set` resp. use them as keys in `std::map`, so the order shouldn't matter -- should it?

fingolfin:
So there still should be a few more tests, but I am pretty confident that all of the code is sufficiently covered by the Normaliz test suite, and/or is trivial enough. And then the Boost code should be ripped out.

Most importantly, the `BUILDSYSTEM=autotools-makedistcheck` Travis job needs to be fixed, but I don't quite understand right now what's the problem. If somebody knows, let me know, otherwise I'll investigate tomorrow or so.

w-bruns:
The options -cis were meant for A553 and A543 and lo6.

Please run m13, 14, 15 without -I, because it switches off  FVecor for which the 0-1-vectors are  Crucial.

But even before these tests, I agree to remove boost::dynamic_bitset.

fingolfin:
So just `-cs` ? OK, I'll let the benchmarks run tonight.
```

**Comment by fingolfin** [comment]
OK, without `-i`, things are quite a bit slower, and the advantage less clear, but still there: +5% speed. I reduced the number of warmup runs and repetitions for these measurements, and M15 is still running as I write this.

M13:
```
$ ARGS='-cs playground/m13' ; hyperfine --warmup 1 "source/normaliz $ARGS" "../normaliz-dynamic_bitset/source/normaliz $ARGS"
Benchmark #1: source/normaliz -cs playground/m13
  Time (mean ± σ):     29.900 s ±  0.250 s    [User: 29.714 s, System: 0.167 s]
  Range (min … max):   29.716 s … 30.589 s    10 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark #2: ../normaliz-dynamic_bitset/source/normaliz -cs playground/m13
  Time (mean ± σ):     28.414 s ±  0.204 s    [User: 28.189 s, System: 0.168 s]
  Range (min … max):   28.212 s … 28.948 s    10 runs

Summary
  '../normaliz-dynamic_bitset/source/normaliz -cs playground/m13' ran
    1.05 ± 0.01 times faster than 'source/normaliz -cs playground/m13'
```

M14:
```
ARGS='-cs playground/m14' ; hyperfine --warmup 1 "source/normaliz $ARGS" "../normaliz-dynamic_bitset/source/normaliz $ARGS"

Benchmark #1: source/normaliz -cs playground/m14
  Time (mean ± σ):     210.616 s ±  2.329 s    [User: 209.357 s, System: 0.815 s]
  Range (min … max):   207.688 s … 215.601 s    10 runs

Benchmark #2: ../normaliz-dynamic_bitset/source/normaliz -cs playground/m14
  Time (mean ± σ):     198.520 s ±  1.330 s    [User: 197.651 s, System: 0.778 s]
  Range (min … max):   196.514 s … 200.302 s    10 runs

Summary
  '../normaliz-dynamic_bitset/source/normaliz -cs playground/m14' ran
    1.06 ± 0.01 times faster than 'source/normaliz -cs playground/m14'
```

M15:
```
ARGS='-cs playground/m15' ; hyperfine --warmup 1 "source/normaliz $ARGS" "../normaliz-dynamic_bitset/source/normaliz $ARGS"

TODO
```

w-bruns:
Still looking good!

fingolfin:
From my point of view, this is ready and could be merged now (assuming the tests pass). I'd prefer if there were more unit tests, but again, I think the stress testing by the Normaliz test suite should already be giving a pretty good coverage.

w-bruns:
I agree completely.

fingolfin:
@w-bruns Tests *did* pass :-)

w-bruns:
I will merge this evening.

w-bruns:
There is a problem: when we change the bitset type in Normaliz, then we should (more or less) simultaneously update PyNormaliz since the face lattice and the incidence matrix are transferred using boost::dynamic_bitset.

fingolfin:
See https://github.com/Normaliz/PyNormaliz/pull/67

fingolfin:
Also NormalizInterface will need to be updated, I'll apply a fix there. I'll make a new release of that once 3.8.1 is out.

fingolfin:
@w-bruns I've updated this one last time (I hope), to make the PyNormaliz transition smooth. To this end, I rewrote the helpers `popcount` and `limb_find_first`. Why? Well, the code in there was relying on preprocessor macros `SIZEOF_LONG_LONG`, `HAVE___BUILTIN_POPCOUNTLL`, `HAVE___BUILTIN_CTZLL`. But in PyNormaliz, those were not set, and that triggered an `#error` directive. I now rearranged the code so that it checks for `HAVE___BUILTIN_CTZLL` and `HAVE___BUILTIN_POPCOUNTLL`; and if that is not present, it safely falls back to the generic code. I also added in 32bit code, just in case somebody wants to use 32bit instead of 64bit code (by changing `limb_t` from `unsigned long` to say `unsigned int`). Doing so may require a few more changes, though, but I figured it couldn't hurt (we'll see if I was right about that...)

w-bruns:
I have a made a branch last_boost from master that I will push to GitHub in a few minutes in order to preserve the boost version.

w-bruns:
Please resolve the merge conflict in .gitignore

PR #272, opened 2019-09-18
Add unittests for dynamic_bitset (DO NOT MERGE)

fingolfin:
This uses the single-header testing framework Catch2, see
<https://github.com/catchorg/Catch2>.

The idea is that these tests shall pass with both boost::dynamic_bitset,
but also our own dynamic_bitset.

PR #270 contains this, and the two PRs are developed in parallel, so please don't merge this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,6 +84,7 @@ autom4te.cache
 /config.sub
 /configure
 /missing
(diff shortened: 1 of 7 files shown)


fingolfin:
This has served its purpose.

PR #274, opened 2019-09-18
configure.ac: put generated files into cnf subdir

fingolfin:
Let's see if this works on its own (used to be part of PR #271)


w-bruns:
I think it is a good idea to put the files generated by the configuration process into a separate directory. They have annoyed me several times already, and I am never sure what I am allowed to delete.

Unclear to me why one build failed.  It is the oldest Xcode version we are testing.

fingolfin:
I am still looking into why that Xcode 6.4 builds. There is another build using Xcode 9.2 but otherwise configured identically, which works. I am also seeing some weird error messages in there:
```
...
+./bootstrap.sh
/usr/local/bin/glibtoolize: line 406: /usr/local/Library/ENV/4.3/sed: No such file or directory
/usr/local/bin/glibtoolize: line 2513: /usr/local/Library/ENV/4.3/sed: No such file or directory
...
```
This makes me wonder if perhaps the homebrew install there is too old / buggy...

And of course Travis is printing a big banner above those jobs ([see e.g. here](https://travis-ci.org/Normaliz/Normaliz/jobs/586782372)_, which says: "Xcode 6.4 in Travis CI is deprecated and will be removed in January 2019". So I guess it's a miracle those builds even still work today ;-).

So perhaps the best solution is to just drop that job. There is not much we gain from testing with that old version anyway.

fingolfin:
Specifically, I wonder if the above error means that the GNU libtool installed via homebrew relies on GNU sed installed via homebrew, but due to a missing dependency, GNU sed was not installed. If so, I can probably work around the issue in one of several ways (get hombrew to install the missing GNU sed; patch libtool to use another copy of GNU sed; install a fresh libtool; etc.). But I don't think it's worth the hassle, I'd just drop that redundant job. Besides being deprecated, the macOS job runners are also a scarce commodity, slowing down the turn around times of the Normaliz Travis builds: only 2 macOS jobs are run in parallel, but 5 Linux jobs are run parallel. In a Normaliz Travis build, there are currently 8 Linux job (and 8/5=1.6), and 6 macOS jobs (6/2 = 3).

w-bruns:
I think we can drop Xcode 6.4.

fingolfin:
@w-bruns done, tests pass now

PR #280, opened 2019-09-20
Treat ClassGroup as a complex structure, not a vector

fingolfin:
While technically, ClassGroup is implemented as std::vector<Integer>, it is
not a "true vector". Thus getVectorConeProperty() does not support it.

This change ensures that output_type(ClassGroup) returns OutputType::Complex
instead of OutputType::Vector, so that code using the generic property access
APIs does not end up trying to call getVectorConeProperty(ClassGroup).

This is an alternative to PR #242. While I still think that PR is superior, I need this situation to be resolved one way or another, so this PR is better than nothing for me.

Resolves #242.

### DIFF
--- a/source/libnormaliz/cone_property.h
+++ b/source/libnormaliz/cone_property.h
@@ -92,7 +92,6 @@ namespace ConeProperty {
         Dehomogenization,
         WitnessNotIntegrallyClosed,
         GeneratorOfInterior,
-        ClassGroup,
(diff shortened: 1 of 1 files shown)


w-bruns:
Now I see the problem ... Sorry that I did not realize it before. There is a contradiction between cone_property.h and cone.cpp.

I think ClassGroup should really be treated as a complex structure, even if it is syntactically a vector. But it is not a vector in the sense of geometry or algebra.

PR #283, opened 2019-10-11
Don't overlink

saraedum:
e-antic's C++ interface only depends on GMP(xx) and FLINT. (and arb through renf_elem.h)

see https://github.com/conda-forge/normaliz-feedstock/issues/10.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -251,7 +251,7 @@ AC_ARG_ENABLE([enfnormaliz],
     [enable_enfnormaliz=check])
 AS_IF([test "x$enable_enfnormaliz" != xno],
     [AC_MSG_CHECKING([whether e-antic headers and library are available])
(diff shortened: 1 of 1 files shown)


w-bruns:
As the failed Travis builds show, there is no overlinking. arb is crucially needed by e-antic, and MPFR is needed by Flint, even if one does not do any floating point computations in it.

saraedum:
I had missed that reference https://github.com/videlec/e-antic/blob/master0/e-antic/renf_elem.h#L17.

The question is not really whether it is used but whether the usage is exposed.

Let's see whether mpfr is really used in that way.

w-bruns:
Well, we can omit -lmpfr when it comes to e-antic. But I think this is only possible because -lmpfr is linked with Flint. I think we can even omit -lflint in connection with e-antic. 
It must in any case be guaranteed that Flint is available when e-antic is asked for.

PR #287, opened 2019-10-29
Remove references to Qnormaliz, nmzIntegrate; fix typos

fingolfin:
Also remove some dead code

In normaliz.tex, there is still a reference to nmzIntegrate on Windows which says, among other things, that "NmzIntegrate can still be used (in all operating systems)", which I think is at best misleading? But I don't know what really is the state of things on Windows, so I didn't touch that paragraph.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,8 +7,6 @@ norm64*
 normbig*
 normaliz
 normaliz-*
-nmzIntegrate
-Qnormaliz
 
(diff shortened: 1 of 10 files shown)


w-bruns:
I have already removed that paragraph, but in my current development branch. So there is no need to do anything. Meanwhile there are enough possibilities to run the Linux version under Windows.

PR #288, opened 2019-10-29
Better GMP_INSTALLDIR, CFLAGS, CPPFLAGS, LDFLAGS handling

fingolfin:
- don't force static linking GMP and CoCoA
- fix GMP_INSTALLDIR for e-antic
- don't override {C,CPP,LD}FLAGS provided by the user

With this, I can actually use these scripts on my computer :-)


fingolfin:
This PR does not modify C++ sources, so it can be merged independently from PR #290.

PR #289, opened 2019-10-29
Cleanup processInputPolynomial

fingolfin:
Fix indentation, and change a for loop to modern ranged for loop;
note that since G is a reference, we can remove one assignment
to factorsRead[i] in the loop.

The indentation was odd (not to say: broken) because apparently code from different places with completely different formatting styles was mixed. Which makes me wonder: @w-bruns do you have some kind of "preferred formatting style"? Right now, I am just trying to match whatever the surrounding code looks like, but that also is quite inconsistent. I personally prefer try to format code in a somewhat uniform way (but don't care much which way that is, exactly, as long as it doesn't change every 3 lines :-). Perhaps you'd be interested in adding a `.clang-format` file, and then using [clang-format](https://clang.llvm.org/docs/ClangFormat.html) resp. `git clang-format` to format new code (and one could even consider reformatting the existing code -- but I know this can be highly controversial ;-) just wanted to mention the possibility at least once).

### DIFF
--- a/source/libnormaliz/nmz_polynomial.cpp
+++ b/source/libnormaliz/nmz_polynomial.cpp
@@ -603,66 +603,65 @@ RingElem processInputPolynomial(const string& poly_as_string, const SparsePolyRi
   if(verbose_INT)
     verboseOutput() << "Polynomial read" << endl;
   
-  homogeneous=true;                                     // we factor the polynomials read
-  for(i=0;i< (long) factorsRead.size();++i){ // and make them integral this way
-                                             // they must further be homogenized
-                                             // and converted to polynomials with ZZ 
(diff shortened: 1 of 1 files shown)


w-bruns:
I am afraid that I have changed my formatting habits several times. It would indeed be good to have a uniform indentation, for example.

fingolfin:
So, you could try to play with https://zed0.co.uk/clang-format-configurator/ to create a config that suites your needs. Start at the top, to select whichever "base style" you like best, then tweak it as you like. Then "save" the result, and send it to me?

fingolfin:
@w-bruns if you are OK with merging this PR, I'd then update PR #290

PR #290, opened 2019-10-29
clang-format all sources

fingolfin:
Based on initial clang-format settings by @w-bruns I made this PR to demo the effect of a code format. Recall that you can also view a PR diff with "ignore whitespace" ([click here to directly get to that](https://github.com/Normaliz/Normaliz/pull/290/files?utf8=✓&diff=unified&w=1))


w-bruns:
I think we should use this formatting once the merge conflicts have been resolved and the TRavis tests are successful.

I have just pushed master to GitHub again. My last development branch is merged into it.

fingolfin:
Yeah, sure. The last commit can be regenerated with this command:
```
git ls-files -- "*.cpp" "*.h" | xargs clang-format -i && git ci -a -m "clang-format all sources"
```

.clang-format
```diff
@@ -0,0 +1,7 @@
+BasedOnStyle: Google
+ColumnLimit: 130
```
fingolfin: @w-bruns I chose this number semi-randomly: the default is 80, I think but my feeling is that this is a bit low on modern systems, where one has much bigger screens. On the other hand, I feel that some limit is needed, else some stuff gets hard to read. But I am happy to change this something else, if you want it

source/libnormaliz/HilbertSeries.h
```diff
@@ -323,30 +313,29 @@ class IntegrationData{
...
 
     const vector<mpz_class>& getCyclotomicNum_ZZ() const;
     // returns the denominator, repr. as a map of the exponents of the cyclotomic polynomials
     const map<long, denom_t>& getCyclotomicDenom() const;
-    
+
     void computeWeightedEhrhartQuasiPolynomial() const;
     bool isWeightedEhrhartQuasiPolynomialComputed() const;
-    const vector< vector<mpz_class> >& getWeightedEhrhartQuasiPolynomial() const;
+    const vector<vector<mpz_class> >& getWeightedEhrhartQuasiPolynomial() const;
```
fingolfin: @w-bruns Note that we could add the option `Standard: Cpp11` to get this instead:
```suggestion
    const vector<vector<mpz_class>>& getWeightedEhrhartQuasiPolynomial() const;
```
Let me know what you prefer.

source/libnormaliz/automorph.cpp
```diff
@@ -189,601 +187,566 @@ bool AutomorphismGroup<renf_elem_class>::make_linear_maps_primal(const Matrix<re
...
-            LinFormPerms=result.LinFormPerms;
-            LinFormOrbits=convert_to_orbits(result.LinFormOrbits);
-        }
-        else{
-            //linform_data_via_lin_maps(); 
+    if (LinFormsRef.nr_of_rows() > 0) {
+        if ((method == AutomParam::E || method == AutomParam::G) && !using_renf<Integer>()) {
+            LinFormPerms = result.LinFormPerms;
+            LinFormOrbits = convert_to_orbits(result.LinFormOrbits);
+        } else {
```
fingolfin: Which `else` style do you prefer?

.clang-format
w-bruns: I think 130 is o.k.

source/libnormaliz/HilbertSeries.h
w-bruns: I would not mind >>. However, do we run into problems with older compilers? We could leave > > for a while.

source/libnormaliz/automorph.h
```diff
@@ -37,77 +37,80 @@ using namespace std;
...
 
-template<typename Integer> class Cone;
-template<typename Integer> class Full_Cone;
-template<typename Integer> class Isomorphism_Classes;
+template <typename Integer>
+class Cone;
+template <typename Integer>
+class Full_Cone;
+template <typename Integer>
+class Isomorphism_Classes;
```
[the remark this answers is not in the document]
w-bruns: Hard to tell. I am used to have templates in a line by their own, as introduced by Christof.

source/libnormaliz/automorph.cpp
w-bruns: Good!

source/libnormaliz/HilbertSeries.h
fingolfin: Sure, let's leave it for now; I added the option but commented it out

fingolfin:
Travis failed because the #includes got sorted; I fixed this for now by disabling `SortIncludes`. I think on the long run, it would be beneficial to switch that back on, but it'll require some changes; in particular, the practice of `#including` .cpp files caused problems. Perhaps we can stop doing that at some point... :-)

w-bruns:
I have now pulled master with the new formatting and looked at the files more closely. 

There is one change that I am not convinced of: the insertion of blanks around **all** binary operators and binary relations. I personally prefer a more compact style, with spaces where they seem appropriate (to me).

w-bruns:
Can we reformat as suggested in my last comment? I can of course do it with sed.

fingolfin:
@w-bruns sorry, didn't see your comments on this closed PR.

As a matter of fact, I don't think clang-format supports that at all. So if you don't want spaces around binary operators, one needs to find another code formatter, or not use a code formatter at all.

(I am also not quite sure what exactly you want in the end, as you seem to say that you want blanks in *some* cases, just not all, which doesn't sound like something that could be automated).

PR #291, opened 2019-10-29
Enable Codecov support

fingolfin:


fingolfin:
I've now rebased this. But I am not sure what you want me to do? I disabled the coverage for two of the tests, in one case because it is irrelevant for `makedistcheck` but causes interference (the `.gcov` files upset the distcheck), in the other case because the build already is close to the 50 minutes limit imposed by Travis, and making it just a little bit slower is enough to make it exceed the time limit sometimes.

Of course as  you can see, it is trivial to disable the test in other cases, but I don't really see the need to do so? But if you want me to disable it in other cases, just tell me which. Or, just merge the PR, then disable those you don't like?

w-bruns:
My suggestion aimed at the reduction of system load on Travis. If there is one build whose code coverage contains all the others, then this run should be enough.

Will we finally see which lines in the files are covered?

fingolfin:
You should already now be able to see which lines in each file are covered, see e.g. <https://codecov.io/gh/Normaliz/Normaliz/tree/master/source/libnormaliz> or <https://codecov.io/gh/Normaliz/Normaliz/src/master/source/libnormaliz/full_cone.cpp>.

fingolfin:
However, I just noticed that coverage info is somewhat off, because many templates which are not instantiated are for some reason not counted at all.

fingolfin:
For example, there is tons of really dead code in https://codecov.io/gh/Normaliz/Normaliz/src/master/source/libnormaliz/vector_operations.h which never is called, or only is called from other dead code. I did some googling, but it seems there is no easy way to get them marked as code by tools like `gcov`. But then it's still easy to spot this "dead code", the only problem is that it is not counted towards the "uncovered code", which is a bit annoying, but not horribly so.

w-bruns:
Let us leave 3.8.2 as it is now. We will clean up in 3.8.3.

fingolfin:
Oh sure. Also, I am don't quite trust the results for `vector_operations.h`. I verified for a few of the templates which are "white" (i.e., not even code was generated), and they were indeed unused; but I just found one, `order_by_perm`, which *is* called from code that is executed, see e.g. <https://codecov.io/gh/Normaliz/Normaliz/src/master/source/libnormaliz/matrix.cpp#L3474>. Similar for `conjugate_perm` which is used here: <https://codecov.io/gh/Normaliz/Normaliz/src/master/source/libnormaliz/cone.cpp#L4345>.

This is somewhat strange. I'll look into it.

PR #292, opened 2019-10-29
Reduce number of Travis builds for faster CI turnaround times

fingolfin:
- remove autotools-enfnormaliz-nauty build on Linux with clang:
  we already test this config on macOS with clang
- remove macOS tests with xcode8.3 and xcode9.2: they add little benefit
  (if Normaliz compiles with xcode7.3 and xcode10.2 then most likely with
  these intermediate Xcode versions, too), but they cost us severely in CI
  runtime, as we only can run 1-2 macOS jobs in parallel on Travis
- drop the `-makedistcheck` from the xcode7.3 build: it had no effect


w-bruns:
This is o.k. But we should have a build on Linux that draws all optional libraries (did not exist so far on Linux, but on Mac). I will  extend one of the builds.

PR #294, opened 2019-11-01
Reduce code duplication in cone constructors via templates (WIP)

fingolfin:
One caveat about this is that it required me to change a few calls to Cone constructors, because with this change those calls became ambiguous. I am slightly worried that this could affect other client code. I think part of the problem is that we have automatic conversion from `Matrix<T>` to `vector<vector<T>>` and vice versa. 

Also, I get a linker error I don't understand right now. I don't have time right now to debug this, but wanted it to not get lost, so I put this here as a WIP PR


w-bruns:
Once I tried to make the number type a template and failed. I don't remember why.

Could we (first) reduce the number of variants by dropping the versions with vector<vector<...> > completely? This is something I want to try, but haven't done it yet. The automatic conversions should allow this.

As far as I can see, the linker error occurs only for maxsimplex (an example program for libnormaliz). The linker cannot find some specializations of the templated constructors. I think the specialization used by maxsimplex does not occur in Normaliz anywhere.  Perhaps we have to force their creation in cone.cpp explicitly.

fingolfin:
Ahh, that makes sense. I'll get to it.

As to reducing the number of constructors by only support Matrix: Of course I can also do that, if that is OK by you.

w-bruns:
Is of course o.k.!

w-bruns:
The problem on Travis occurs with older compilers as I have now seen. 
I have just pushed the branch cone_construction to Travis. It realizes this pull request, expect that I have replaced the insertions into the maps by assignments using [ ]. This seems to go through. Therefore this pull request can be closed.

w-bruns:
One more remark: for the further simplification of implementing only the Matrix version and to trust in implicit conversions can be deferred until the input processing in  cone.cpp is changed to Matrix.

w-bruns:
All Travis builds o.k. Therefore I close this PR

PR #296, opened 2019-11-19
Travis: disable optimizations for one job ... (WIP, DO NOT MERGE)

fingolfin:
... so that code coverage for functions which would otherwise by inlined is emitted.

---

I am actually not sure if that's realistic or just nonsense, so consider this an experiment... once tests have run, we can inspect what codecov made of it and decide based on that whether it's worthwhile or not.


fingolfin:
Unfortunately this had zero effect.

PR #300, opened 2019-11-24
Refactor configure.ac

### LINKED ISSUE #233: configure should error when e-antic is not available and --enable-enfnormaliz is specified
Currently, if dependencies are not met (ie flint and e-antic) the following configure invocation

    ./configure --enable-flint --enable-enfnormaliz

does not error. The outcome is that Normaliz will be built without enfnormaliz support.

### LINKED ISSUE #257: Confusing ./configure --help
The configure help is somewhat misleading:
```
  --enable-flint[=ARG]    enable flint [default=check]

Optional Packages:
  --with-PACKAGE[=ARG]    use PACKAGE [ARG=yes]
  --without-PACKAGE       do not use PACKAGE (same as --with-PACKAGE=no)
  --with-flint=DIR        provide location of flint installation prefix or
                          source directory
```

So setting `--without-flint` does not disable the FLINT *dependency* but sets the directory where FLINT can be found to `no`.

It would be nice if this would be closer to common practice, i.e., `--with-flint` forces to build with FLINT dependency, `--with-flint=prefix` builds with FLINT from `prefix` and `--without-flint` disables FLINT (probably the default here.) `--enable/--disable` would go away here because this is feature is fully controlled by a dependency.

This would be of course be a change that breaks some build scripts so it's probably too late for such a change.

fingolfin:
This is a work-in-progress alternative to PR #260. It is unfinished, but I wanted to put it here so it's not forgotten. I simply run out of time finishing it, but will hopefully be able to change that soon.

Once finished, this resolves #257 and resolves #233.

### DIFF
--- a/.travis-build.sh
+++ b/.travis-build.sh
@@ -13,37 +13,45 @@ NMZDIR=${PWD}
 NMZ_OPT_DIR=${PWD}/nmz_opt_lib
 INSTALLDIR=${NMZDIR}/local
 OPTLIBDIR=${INSTALLDIR}/lib
+export NMZ_COMPILER=$CXX
+
+if [ "x$NO_OPENMP" != x ]; then
+    CONFIGURE_FLAGS="${CONFIGURE_FLAGS} --disable-openmp"
+fi
 
 # install dependencies
 case $BUILDSYSTEM in
-    *-nauty*)
-        export NMZ_COMPILER=$CXX
+    *nauty*)
         ./install_scripts_opt/install_nmz_nauty.sh
+        CONFIGURE_FLAGS="${CONFIGURE_FLAGS} --with-nauty=${INSTALLDIR}"
         ;;
 esac
 case $BUILDSYSTEM in
-    *-flint*)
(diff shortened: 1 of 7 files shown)


fingolfin:
This is now hopefully close to being finished. We'll see once the Travis tests pass. Then, I should do another round of tests as to whether it really works as intended; i.e.,
- `./configure`    # with none of the deps visible in *FLAGS
- `./configure`    # with som of the deps visible in *FLAGS
- `./configure`    # with all of the deps visible in *FLAGS
- `./configure --without-FOO`    # repeated for each FOO individually, with all/some/none visible in *FLAGS
- `./configure --without-FOO1 --without-FOO2` ...   # disable all deps
- `./configure --with-FOO`   # with none of the deps visible
- `./configure --with-FOO=PATH`   # with none of the deps visible
- `./configure --with-FOO1=PATH --with-FOO2=PATH ...`   # with none of the deps visible

w-bruns:
Very good. I will have a close loom and test at home.

PR #304, opened 2019-11-27
Fix .codecov.yml syntax

fingolfin:
This fixes a validation error reported by `yamllint`. Perhaps this will make the config actually work...


fingolfin:
I see no effect here, but perhaps this must be merged first. I think there is no harm in it, the worst thing that can happen is that nothing happens...

PR #311, opened 2019-12-06
Remove semistatic build, including NMZ_SHARED, save_dynamic

### LINKED ISSUE #307: Error running install_normaliz_with_eantic.sh
For me, running `install_normaliz_with_eantic.sh` dies in an error like this (I assume other `install_normaliz*.sh` variants are affected as well, but I didn't try):

```
...
OPT_LIB_DIR

installing with --disable-shared
Found see next line
true
Hiding Linux
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
configure: error: source directory already configured; run "make distclean" there first
```

This leaves things in a broken state, as shared libraries were moved to `local/lib/save_dynamic/` and remain there.

fingolfin:
Resolves #307

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -36,7 +36,6 @@ doc/Normaliz.pdf, Section 11.2.
 They installation can in addition be controlled by several environment variables:
 
 NMZ_COMPILER: overrides the standard choice of compiler (g++ on Linux, clang++ on Mac)
-NMZ_SHARED: if set, only shared libraries will be built; binaries depend on all of them.
 NMZ_PREFIX: if set defines the path for make install; 
     by default it is "local" in the Normaliz directory
 GMP_INSTALLDIR: sets the path to the GMP files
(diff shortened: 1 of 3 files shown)


w-bruns:
I think libnormaliz.a is no longer produced . If so, it cannot be longer copied to libnormaliz (see end of install_normaliz.sh). It is not needed. No problem if the copying is removed.

w-bruns:
I think we must also remove NMZ_MAC_STATIC. This was meant to produce an almost static binary for Mac OS (as is done by the Travis builds of the master branch). But this requires the "hiding" of the dylib files which is no longer done. Anyway, we should look into the matter.

PR #312, opened 2019-12-09
Refactor Travis integration

fingolfin:


w-bruns:
Not clear to me whether all tests are done that should be done for the respective configuration. For example, with "eantic cocoa nauty" all tests should be run, but I could only see the tests of basic Normaliz.

w-bruns:
The "hiding" of dynamic libraries should depend on "static", not on "eantic" (even if the "static" is combined with "eantic" at present.).

The same for the compiler from HomeBrew.

Nevertheless not clear to me why the tests for e-antic are not done. Those for nauty and CoCoALib are done. Is

BUILDSYSTEM=eantic-cocoa NO_OPENMP=yes

built with e-antic? I see

checking whether e-antic headers and library are available... no

in the log file.

Is it o.k. to ask for "-enatic" if if BUILDSYSTEM starts with "eantic" ?

fingolfin:
Just to clarify: the purpose of this PR was / is just to "move things around". Any behaviour change is unintended. I agree that more changes should be done, along the lines you suggest; this PR was meant to "lay the groundwork" for that, by first cleaning up everything to make it easier to reason about. I'll investigate what went wrong.

fingolfin:
OK I think I got it: removing the prefix `autotools-` from the strings in `BUILDSYSTEM` meant that e.g. `*-eantic` did not match for `BUILDSYSTEM=eantic-nauty-cocoa`. Let's see if that fixes it.

PR #320, opened 2020-01-09
Overhaul Travis setup: install prebuilt dependencies, use more Travis phases

fingolfin:
<s>Let's see if we can figure out what causes the recent Travis failures.</s>

This PR now morphed into one that tries to overhaul the Travis setup:
- install dependencies via apt resp. homebrew were possible, to reduce compilation time
- move calls to `install_*.sh` scripts out of `.travis-build.sh` into `.travis.yml`; that way, we have full control over which of them to call when; furthermore, the resulting Travis logs are much more useful (each `install_*` now gets its own folded section, with timing information)
- unified a lot of things in the Travis setup and the install scripts (this may lead to a few regressions initially, but I am confident we can fix any of those quickly)


fingolfin:
@w-bruns as you can see on https://travis-ci.org/Normaliz/Normaliz/jobs/634790718?utm_medium=notification&utm_source=github_status the problem (at least in that that job) is caused by a download failure, caused by a failure to DNS resolve `ftp.gnu.org`. We actually have experienced similar issue with `www.gap-system.org` for the past year. We suspect it is an issue with Travis' infrastructure, but are not sure so far. We usually restart the failed jobs. However, this may not be the only problem, and things here get substantially complicated by the fact that we pipe *all* build output to `/dev/null` for these dependencies, which makes debugging a nightmare. I'll look into changing this so that we only pipe the output of the final `make` of each 3rd party install to `/dev/null` (or perhaps even some logfile). 

But perhaps more is going on. Hard to say. Maybe even the fact that we redirect output to `/dev/null` is part of the problem, somehow...

w-bruns:
I piped the output to /dev/null in order to have shorter log files.

w-bruns:
By the way, if one wants to reduce the garbage in the Travis logs, one must also pipe stderr to /dev/null. I have introduced in extended_tests: &> /dev/null. I should have removed it.
Nevertheless it is strange that the problem happens only in Mac builds.

fingolfin:
I am fully aware of why the redirect to `/dev/null` must be there: to avoid the log from getting to big, as Travis will abort the build if that happens. But there should be a middle ground possible. Also, it would help to use Travis "grouping" feature, so that we can hide those install logs by default, and focus on the build output for Normaliz itself, and also the output of the test suite. Unfortunately, right now all of that is mixed together. 

Also note how the issue did *not* happen on this PR, for none of the builds...

fingolfin:
@w-bruns OK this is slowly getting somewhere, but there are a few kinks still to be resolved. In particular, I discovered that the recent changes to improve code coverage lead to a regression where building Normaliz with Flint but without e-antic support lead to linker errors. I've added a commit to this PR to fix that.

While this PR is not yet quite ready, I expect to have it there soon. Feel free to peek at it already now, and to ask any questions :-)

fingolfin:
@w-bruns This now passes all tests. I am afraid the speed up is not very big (from about 2 hours 20 minutes down to perhaps 2 hours 10 minutes; the numbers fluctuate quite a bit, though, so we'll have to see if this holds up).

But a major improvement is that at least now one can see how long compiling each dependency takes. E.g. for the slowest build (number 7, on macOS), looking at [the Travis logs](https://travis-ci.org/Normaliz/Normaliz/jobs/640169882?utm_medium=notification&utm_source=github_status) we can see that out of ~45 minutes, we spend ...
- ~1 minutes updating Homebrew;
- ~5 minutes installing LLVM;
- ~2 minutes compiling mpfr;
- ~12 minutes (!) compiling flint;
- ~8.5 minutes compiling arb
- ~5 minutes for CoCoA.

Now, homebrew has packages for mpfr and flint; installing that should those just a few seconds, and would save us 14 minutes. The reason I am not (yet??) doing that is that it might break the main (?) purpose of that particular Travis job, which is to build static macOS binaries that run "everywhere": To achieve that, this job "hides" the dynamic libraries for flint etc., before linking Normaliz. But it only "hides" dynamic libraries in our own `local` dir, not the ones installed by Homebrew!

I can try to resolve this, e.g. by trying to remove the `dylib` files from the Homebrew `lib` directory. But there might be other consequences of that, so I'll put it into a different commit or even a different PR.

Hmm, now that I think about it, I wonder how those static binaries even work right now, given that we do *not* seem to build our own GMP, but rather use one installed by Homebrew? Perhaps I am missing something...?

fingolfin:
I've now added a commit which tries to install flint&mpfr on macOS via homebrew while still static linking. And I also take care of ensuring gmp is linked statically: my suspicion was right, it wasn't before! In fact it turns out that also `libc++` is linked dynamically -- but the version from the custom LLVM we install, not the system `libc++`. Thus I don't think those static binaries actually run on every Mac... which makes me wonder, are they tested by anybody resp. by some other automation?

fingolfin:
Builds now and is faster, but I messed up the static linking for macOS:
```
+otool -L /Users/travis/build/Normaliz/Normaliz/local/bin/libomp.dylib /Users/travis/build/Normaliz/Normaliz/local/bin/normaliz
/Users/travis/build/Normaliz/Normaliz/local/bin/libomp.dylib:
	@loader_path/./libomp.dylib (compatibility version 5.0.0, current version 5.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.200.5)
/Users/travis/build/Normaliz/Normaliz/local/bin/normaliz:
	/usr/local/opt/flint/lib/libflint-13.5.2.dylib (compatibility version 13.5.0, current version 13.5.2)
	/usr/local/opt/mpfr/lib/libmpfr.6.dylib (compatibility version 7.0.0, current version 7.2.0)
	/usr/local/opt/gmp/lib/libgmpxx.4.dylib (compatibility version 10.0.0, current version 10.2.0)
	/usr/local/opt/gmp/lib/libgmp.10.dylib (compatibility version 14.0.0, current version 14.2.0)
	@loader_path/./libomp.dylib (compatibility version 5.0.0, current version 5.0.0)
	/usr/local/opt/llvm/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.50.4)
```
Will debug tomorrow

w-bruns:
I think the speed up is substantial on Mac.
It would be enough to produce the static Mac version in the master branch. Actually, the deployment is only done in this branch. On the other hand, this does not help much for the code.

fingolfin:
Producing the static version is not what eats the time, though.

fingolfin:
Ah, found the cause for the broken static linking: I introduced a new env var `BREWDIR` but forgot to initialize it *ahem*. Let's see if it works now.

fingolfin:
That fixed it, now I am seeing this in the build log:
```
+otool -L /Users/travis/build/Normaliz/Normaliz/local/bin/libomp.dylib /Users/travis/build/Normaliz/Normaliz/local/bin/normaliz
/Users/travis/build/Normaliz/Normaliz/local/bin/libomp.dylib:
	@loader_path/./libomp.dylib (compatibility version 5.0.0, current version 5.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.200.5)
/Users/travis/build/Normaliz/Normaliz/local/bin/normaliz:
	@loader_path/./libomp.dylib (compatibility version 5.0.0, current version 5.0.0)
	/usr/local/opt/llvm/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.50.4)
```
Which indicates that indeed flint, mpfr etc. are all statically linked. But `libc++` is not, as I mentioned -- and that wasn't the case before either.

However, I can try to fix that as well.

fingolfin:
Nope, removing `libc++.dylib` just results in linker errors. So I reverted this. At least we are not worse off than before.

@w-bruns assuming the tests pass again this time, I think this can be merged.

w-bruns:
As I just have seen, we get the hint

`OMP: Info #270: omp_set_nested routine deprecated, please use omp_set_max_active_levels instead.`

on Mac with the homebrew setup (not on Linux, though, neither with gcc nor with clang). I will check the matter.

w-bruns:
Unrelated question: do we really need two versions of codecov reports (or checks) ? Especially the first one often fails.

w-bruns:
For me everything looks good. If you agree, then I will merge.

fingolfin:
Please merge.

Re: Codecov: I don't understand it. Codecov seems to ignore the `.codecov.yml` settings for this, which say:
```
  status:
    project: false
    patch: false
    changes: false
```
I did verify that they parse it right (see <https://codecov.io/gh/Normaliz/Normaliz/settings/yaml>), but it has no effect. Problem is: their documentation for this file is basically useless....

... or so it was until a few weeks ago. I just had a fresh look, and they massively improved it (I've been complaining to them about this for several years, and had given up hope. Wow). Looking at <https://docs.codecov.io/docs/codecovyml-reference> and <https://docs.codecov.io/docs/commit-status> we might have to change `false` to `off`. I'll try that -- but in a fresh PR. Let's get this one merged.

PR #324, opened 2020-01-27
Do not overlink arb

saraedum:


saraedum:
Last time I had checked, normaliz was not actually using arb.

saraedum:
Actually, not true, you are using arb_get_str() in one place.

PR #335, opened 2020-06-05
Fix #246: OpenMP issue with AppleClang

### LINKED ISSUE #246: Compile Mac OS binary with xCode AND OpenMP
This is indeed possible if one installs the dynamic library for OpenMP from Homebrew (information from Dan Grayson). Some tricks are necessary, but we will do it in the long run.

mahrud:


mahrud:
This is tested on Linux and macOS 10.15 (Catalina), but the solution is not final. On my system, I had to provide the following flag for it to work:
```
OPENMP_CXXFLAGS=-Xclang%20-fopenmp%20-I/usr/local/include%20%20/usr/local/lib/libomp.dylib
```
This is not super easy to make automatic, since it depends on the location of the library on the system.

Note that the fix is slightly different from the fix that Dan suggested in that issue. I use "%20" as the separator (URL encoding of space), he used "%".

mahrud:
Ah, a further change needs to be make to the `libtool` script that is generated by `configure`. Namely:
```diff
diff -Nar -U 5 a/libtool b/libtool
--- a/libtool  2020-06-11 13:23:35.000000000 -0400
+++ b/libtool  2020-06-11 13:23:35.000000000 -0400
@@ -7582,14 +7582,10 @@
       -mt|-mthreads|-kthread|-Kthread|-pthread|-pthreads|--thread-safe \
       |-threads|-fopenmp|-openmp|-mp|-xopenmp|-omp|-qsmp=*)
       func_append compiler_flags " $arg"
       func_append compile_command " $arg"
       func_append finalize_command " $arg"
-      case "$new_inherited_linker_flags " in
-          *" $arg "*) ;;
-          * ) func_append new_inherited_linker_flags " $arg" ;;
-      esac
       continue
       ;;
 
       -multi_module)
       single_module=$wl-multi_module
```

w-bruns:
Thanks for your efforts! I am not yet sure when the next Normaliz release will be published. Hopefully we can take car of this issue then.

fingolfin:
Don't patch generated libtool scripts, that's a very fragile "solution". Rather, should try to understand the underlying issue, and resolve it properly. However, you don't actually explain what the problem is you are trying to overcome... Care to elaborate? Then perhaps I can suggest an alternative solution

I am also very curious about the use of `%20` "URL encoded" characters in your `OPENMP_CXXFLAGS`. That almost certainly is not the appropriate way to handle whatever issue you mean to handle with it... Which again you don't actually describe :-(.

The reference to `/usr/local/lib/libomp.dylib` also doesn't fit with the title of this PR, which suggests that OpenMP with Apple's Clang is supposed to be used, yet `/usr/local/lib/libomp.dylib` is not provided by Xcode. Where does it come from?

mahrud:
@fingolfin I totally agree. This is a terrible fix.

However, the issue is describe in the title: fix #246. The correct compiler flag for openmp on AppleClang is not `-fopenmp`, it's `-Xclang -fopenmp` or `-Xpreprocessor -fopenmp` (not sure if there's a difference). In addition, `libomp` needs to be linked explicitly because, unlike Clang, AppleClang doesn't ship with OpenMP. However, libtool doesn't handle this distinction.

w-bruns:
I am closing this pull request for the time being in order not to merge it inadvertently.  We will come back to it.

mahrud:
This is still an issue, by the way. The configure script doesn't find OpenMP correctly when using AppleClang.

PR #336, opened 2020-06-16
.github/workflows: Add Sage CI workflows

mkoeppe:
This adds a GitHub Actions workflow, testing Normaliz with the new Sage portability testsuite (​see https://doc.sagemath.org/html/en/developer/portability_testing.html, ​​https://researchseminars.org/talk/SageDays109/7/)

On every push of a tag and on every pull request, this creates a "make dist" of Normaliz and then builds a portion of the Sage distribution (the current beta version), including Normaliz and PyNormaliz (from the Normaliz distribution just built) on a large number of different Linux distributions / configurations.

### DIFF
--- a/.github/workflows/ci-sage-linux.yml
+++ b/.github/workflows/ci-sage-linux.yml
@@ -1,0 +1,105 @@
+name: Run Sage CI for Linux
+
+## This GitHub Actions workflow runs SAGE_ROOT/tox.ini with select environments,
+## whenever a GitHub pull request is opened or synchronized in a repository
+## where GitHub Actions are enabled.
+##
+## It builds and checks some sage spkgs as defined in TARGETS.
+##
+## A job succeeds if there is no error.
+##
+## The build is run with "make V=0", so the build logs of individual packages are suppressed.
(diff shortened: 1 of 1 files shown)


w-bruns:
We leave it open.

mkoeppe:
The new workflow will run when you push any tag to GitHub, or when you receive a pull request.

I am working on an extension of it that also includes a build on Cygwin.
Preview: https://github.com/mkoeppe/Normaliz/actions/runs/137292948

Note that there are a number of build failures on old Linux distributions with old compilers, for example on ubuntu-trusty.

w-bruns:
I have made a technical pre-release 3.8.6 in order to trigger the Sage integration builds. Has worked out, but is not yet finished. So far some builds went through, some others failed. More later

w-bruns:
https://github.com/Normaliz/Normaliz/actions/runs/138175811

is the link for the run triggered by my pre-release. Not completely finished yet. The second run caused by the new pull request is at

https://github.com/Normaliz/Normaliz/actions/runs/138583262

PR #345, opened 2020-07-05
configure.ac: drop hardcoded -rpath usage

fingolfin:
Experiment to see what happens if we drop these.

UPDATE: No ill effects visible on Travis, it also seems to work for me locally... so perhaps this could be merged, although there is some risk of regression (but a regression report would still be useful as it might give a clue as to what was meant to be achieved by having these rpath options in there in the first place)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -218,7 +218,7 @@ AS_CASE([$with_flint],
   [check],[],
   [*],[
     FLINT_CPPFLAGS="-I$with_flint/include"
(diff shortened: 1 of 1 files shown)


w-bruns:
Can this PvR be merged?

fingolfin:
I think this can be merged now (I resolved the conflict). There are a few additional failures in the "Sage CI" test runs on macOS, but it looks to me as if the the tests simply failed to run and then timed out after some hours? It's hard to tel, though, as there are no logs, it seems

PR #349, opened 2020-07-07
Drop explicit use of MPIR

fingolfin:
This way, it becomes possible to build Normaliz against GMP on cygwin.
Note that it is still possible to build Normaliz against MPIR, by using
MPIR GMP emulation.

However, this may cause compiler errors on Windows at first, due to `long` vs. `long long` mismatches. I am confident all of these can be resolved, though doing that requires access to a Cygwin system... I might also be able to abuse Julia's [BinaryBuilder](https://binarybuilder.org) for this, to get a VM in which I can build Normaliz. Alternatively, if somebody could just try to build with the patch in this PR and report the errors they see, I can try to suggest patches. But this would be really cumbersome.

Ping @isuruf @saraedum

Resolves #334

PR #354, opened 2020-08-02
CI-sage updates

mkoeppe:


w-bruns:
Before I merge this Pr: why are there 5 failing builds?

fingolfin:
I don't think this is ready for merging?!? @mkoeppe btw you can submit PR as "Draft" so they are clearly marked as such (useful when still tweaking changes in there, as seems to be the case here)

mkoeppe:
The pull request is ready and safe to merge. 

@fingolfin It fixes what you observed in #349 - unrelated testsuite failures in flint on some platforms caused the build to abort.

Unfortunately the Sage ticket that the CI infrastructure is not ready yet. It uses an autotoolized fork of nauty (https://github.com/mkoeppe/nauty) based on Debian's patches, which still has build failures on Linux.

mkoeppe:
I have reverted https://trac.sagemath.org/ticket/27952 to the version that uses upstream nauty.

mkoeppe:
Now I will force-push to this pull-request (without changes) again, just to trigger a re-run of the tests.

fingolfin:
@mkoeppe I appreciate that it was done with good intentions, but I'd rather not be "CCed" by all those SageMath trac items; I am getting a barrage  of (to me) meaningless mails in which I am simply not interested and which I did not request nor agree to receive; and found no way to "unsubscribe" from either. That's normally how I'd describe spam... now again, I understand that this was done only with best intentions, but please unsubscribe me from these trac items again.

mkoeppe:
Lieber Max, I am cc'ing you (and Winfried) on items that are relevant to Normaliz maintenance, which (to my understanding) you have committed to. It works in the same way as GitHub @ mentions that you are sending me. 

When I cc'd you on the new ticket https://trac.sagemath.org/ticket/30271, I should, of course, have commented why I cc'd you and Winfried: You may want to consider using the autotoolized version of nauty so that you can build and use a shared library.

fingolfin:
The difference is that you obviously have an account on GitHub and so can easily "unsubscribe" from these issues if you don't want them, or configure notification settings etc., while I have no account on the SageMath Trac and no visible way to block them. And GitHub doesn't send emails for every trivial change like setting labels etc.. So if I "at" you (which, BTW, I'd immediately stop if you told me you object to it), you get a very small number of emails and also the tools to stop them. I have no such recourse with the Trac mails, and frankly, find it a very questionable design that apparently one can subscribe arbitrary email addresses there. At the very least, an "unsubscribe" link at the bottom of each email could be there (as GitHub sends them).

So, please, don't do that and please, let me know how I can unsubscribe from the existing emails and/or please remove me from them if you can. I am really not interested in autotoolized nauty.

w-bruns:
Please keep me cc on your tickets.

Winfried

fingolfin:
Regarding Normaliz maintenance: I am absolutely committed and willing to work on that, and I think I have established that quite clearly.

However, IMHO the norm in open source is that issues or support requests for project A are reported via the tracker of A; while it is generally not expected that maintainers look through issues of project B to figure out whether between several dozen comments filling many pages there is something that might constitute an issue with A or otherwise be relevant. (Of course if they are willing to do that, as Winfried is, that's fine, but I am already involved with many other projects and don't have the time.)

Point in case: nauty works great over here, both installed via my package manager and via Winfried's scripts. If Sage wishes to use a modified version of it, I fully respect that, but I don't see why that should concern me? Of course if there is a bug in our build system that impedes this, or if there is a change we could make to simplify  things for Sage, I am absolutely willing to work on that! But again, isn't the normal process (which you and other Sage members have followed many times -- thank you for that!) to submit an issue or PR here? Possibly with a comment like "For more info, see also LINK-TO-SAGE-TRAC".

mkoeppe:
Lieber Max,
I know that many of us find these times to be particularly stressful. And I do know how additional emails can add to that. Please take care of yourself. We need you.

But I would kindly ask that we keep discussions to the technical point of the pull request.

mkoeppe:
>  nauty works great over here, both installed via my package manager and via Winfried's scripts.

If you were testing on ubuntu-trusty, you would be seeing the error too.

PR #362, opened 2020-12-18
Some initial GitHub Actions CI tests

fingolfin:
This is *not* the ideal way to set up these things, but rather just a first
step to get our toes wet and figure out what works and what doesn't.
More to follow.

Some things to look into:
- actually make the jobs work
- port over more/all of the Travis jobs
- use a build matrix to reduce duplication among the different jobs (and also to avoid having to pick a "name" for each :-))
- also run Windows tests here
- ... many more things could be done, but one step after the other

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,55 @@
+name: Run tests
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  eantic-nauty-cocoa-extended:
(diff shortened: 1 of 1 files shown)


fingolfin:
BTW I am also happy for others to help out with this (or do it instead of me), if anybody has time and energy for this.

fingolfin:
Err... this should not have been merged, it was not yet working at all (all tests failed)

PR #368, opened 2021-01-22
adapt to e-antic 1.0.0

saraedum:


source/libnormaliz/input.cpp
```diff
@@ -352,7 +352,7 @@ void read_symbolic_constraint(istream& in, string& rel, vector<Number>& left, Nu
         if (coeff_length == 0 || (coeff_length == 1 && coeff_string[0] == '+'))
             coeff = 1;
         if (coeff_length == 1 && coeff_string[0] == '-')
-            coeff = -11;
```
saraedum, line 355: Is that a bug @w-bruns ?

saraedum:
The FLINT is too old in that version of Ubuntu. We need at least groovy gorilla...

source/libnormaliz/input.cpp
w-bruns: Sure, -1 (as mentioned already)

w-bruns:
Three remarks:

1. The Sage tests are not relevant for the success of this pull request. In order to make the situation more transparent, I suggest to rename the file .github/workflows/ci-sage.yml temporarily so that these tests are not run for this pull request. They can be done in the second round.
2. All but one of the Normaliz test builds are successful! The one that fails uses Homebrew LLVM to produce a Mac OS binary with OpenMP enabled.
3. I think we should no longer set the renf in input via the stream. This may have been very helpful at the beginning, but there are better methods now.

w-bruns:
As far as I can see, there is only a problem in the compilation of maxdimplex in the build that fails. This in internal test for libnormaliz. So it should be o.k.

saraedum:
>     1. The Sage tests are not relevant for the success of this pull request. In order to make the situation more transparent, I suggest to rename the file .github/workflows/ci-sage.yml temporarily so that these tests are not run for this pull request. They can be done in the second round.

Done.

>     2. All but one of the Normaliz test builds are successful! The one that fails uses Homebrew LLVM to produce a Mac OS binary with OpenMP enabled.

Actually, a `make installcheck` fails on Ubuntu because it cannot find `libantic.so`. I have to investigate what's going on exactly.

>     3. I think we should no longer set the renf in input via the stream. This may have been very helpful at the beginning, but there are better methods now.

While I fully agree that this should be changed eventually, I think this would introduce too many changes here. The only way I see is to pass around a reference to the renf all the time. This would change the interface in quite a few places.

w-bruns:
O.K. We keep the input stream transfer of the renf for the time being.

As to the Sage tests: I am not sure which version of PyNormaliz is used. The present master of Normaliz defines a data structure for triangulations and similar decompositions that is not backward compatible. The current Normaliz master must be combined with the current PyNormaliz master (labeled 2.14). (And we have not yet updated the disthook in Makefile.am ...)

w-bruns:
Let me add a comment on the "static" build for Mac OS: we don't have a Mac at our disposal. Therefore we use GitHub CI to compile a binary that is as static as possible and make it available as an "artifact". To this end we delete the dynamic libraries that would win against the static ones in the compilation. Unfortunately "allstatic" is not strong enough as we have experienced (we would still link arb.dylib -- no idea why).

w-bruns:
Some checks fail because the precision of the number field now changes -- no surprise. The tests have two levels: STRICT=no (default) and STRICT=yes. Unfortunately we forgot to omit the number field precision from the STRICT=no level.

If this is the only reason why some builds fail, we can safely merge the pull request. We will take care of the issue above.

w-bruns:
In the failing build with "static in its name I see that clang complains about missing dylibs. As already mentioned, these have indeed been removed in order to force the linking of the corresponding static libraries. Why don't these exist or aren't tried? After all, we do not have a dynamic ver4sion of CoCoA and Normaliz is built nevertheless with it.

saraedum:
@w-bruns I had a look at what is happening. The `-all-static` is not used because it has to be there at configure-time. When `make` is called, afaik the LDFLAGS are not looked at anymore.

Even if it's set it's not being used by most libraries because they are not using libtool. One can fake that by creating some .la files but even that does not work because there are duplicate symbols and from a bit of research it appears that macOS cannot be made to accept that. See my latest commit for details.

Why do you need a static build? (How "static" is it actually?)

saraedum:
To the last point. In my experience static builds are not worth the headaches that maintaining them costs. Imho, others have already fixed that problem several times, so I'd recommend to just use some existing solution, e.g., `conda-pack`:

```sh
$ mamba create --quiet --yes -n normaliz normaliz
  Package      Version  Build       Channel                  Size
───────────────────────────────────────────────────────────────────
  Install:
───────────────────────────────────────────────────────────────────

  arb           2.19.0  h13e07f3_0  conda-forge/osx-64     Cached
  e-antic        0.1.8  h451bf31_1  conda-forge/osx-64     Cached
  gf2x           1.3.0  h0959a2e_1  conda-forge/osx-64     Cached
  gmp            6.2.1  h2e338ed_0  conda-forge/osx-64     Cached
  libcxx        11.0.1  habf9029_0  conda-forge/osx-64     Cached
  libflint       2.6.3  haaaf077_0  conda-forge/osx-64     Cached
  llvm-openmp   11.0.1  h7c73e74_0  conda-forge/osx-64     Cached
  mpfr           4.0.2  h72d8aaf_1  conda-forge/osx-64     Cached
  normaliz       3.8.5  h3eef696_1  conda-forge/osx-64     Cached
  ntl           11.4.3  h2295e1f_1  conda-forge/osx-64     Cached

  Summary:

  Install: 10 packages
$ conda pack -n normaliz -o normaliz-3.8.5-macos.tar.gz
Collecting packages...
Packing environment at '/Users/ci/minimamba/envs/normaliz' to 'normaliz-3.8.5-macos.tar.gz'
[########################################] | 100% Completed |  2.3s
$ mkdir -p normaliz-3.8.5
$ tar zxf normaliz-3.8.5-macos.tar.gz -C normaliz-3.8.5
$ ./normaliz-3.8.5/bin/normaliz
ERROR: No project name set!
$ otool -L ./normaliz-3.8.5/bin/normaliz
./normaliz-3.8.5/bin/normaliz:
	@rpath/libnormaliz.3.dylib (compatibility version 12.0.0, current version 12.5.0)
	@rpath/libeanticxx.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	@rpath/libeantic.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	@rpath/libarb-2.10.0.dylib (compatibility version 2.10.0, current version 2.10.0)
	@rpath/libflint-14.dylib (compatibility version 14.0.0, current version 14.0.3)
	@rpath/libgmpxx.4.dylib (compatibility version 11.0.0, current version 11.1.0)
	@rpath/libgmp.10.dylib (compatibility version 15.0.0, current version 15.1.0)
	@rpath/libc++.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
```

w-bruns:
The reason for having an almost static binary is that it is the most user friendly version one can have. The user need not install a compiler. The Mac OS binary that we have produced so far needs only 2 or 3 system dylibs that are standard on every Mac. We have tested this with some users. 

As to all-staic I can say that on Linux it is enough to give this to make as an option. It produces a a fully statically linked binary that runs "everywhere". On MacOS the all-static for make almost worked despite of the presence of the  dylibs, except for arb. After we had seen that we removed the dylibs again before make. That works (see our Travis logs or the new GitHub logs).

We will think about this question again.

saraedum:
> As to all-staic I can say that on Linux it is enough to give this to make as an option. It produces a a fully statically linked binary that runs "everywhere". On MacOS the all-static for make almost worked despite of the presence of the dylibs, except for arb. After we had seen that we removed the dylibs again before make. That works (see our Travis logs or the new GitHub logs).

That's curious. You're right. On my macOS it has no visible effect when passed to make but it is added to the linker flags on Linux. It was my understanding that libtool cannot do static builds without .la files but maybe there's some fallback logic that I am not aware of.

It might also be that since I need versions that are newer than the ones that come with Debian, I build flint/nauty/arb (and now also antic) from the upstream tarball. These might be different than what came with Debian/Brew.

saraedum:
Apart from the static builds which might be beyond the scope of this, I think that I am done here.

w-bruns (commented):
Thanks for your intensive work! I approve the changes, but we would like to merge this pull request into our branch new_we-antic. 
Can you change the destination of the pull request? If not, we will do it before merging.

w-bruns:
I have now fetched new_e-antic to my PC. I run into two problems:

1) Though I have not changed the MPFR source, an attempt to install MPFR results in
```
Checking for mpfr-4.1.0.tar.gz... found existing file, using that
Verifying SHA256 checksum of mpfr-4.1.0.tar.gz...
   valid SHA256 checksum
CDPATH="${ZSH_VERSION+.}:" && cd . && /bin/bash /home/winfried/Dropbox/git_normaliz/nmz_opt_lib/MPFR_source/mpfr-4.1.0/missing aclocal-1.16 -I m4
/home/winfried/Dropbox/git_normaliz/nmz_opt_lib/MPFR_source/mpfr-4.1.0/missing: Zeile 81: aclocal-1.16: Befehl nicht gefunden
WARNING: 'aclocal-1.16' is missing on your system.
```
The newest version I can get via apt (Ubuntu 18-04) is 1.15. What introduces 1.16 to MPFR?
For the time being, I can use the installed MPFR, but the problem must be taken care of.

2) The linker complains
```
/home/winfried/Dropbox/git_develop/local/lib/libflint.a(fmpq_poly_inlines.o): In function `fmpq_poly_add_fmpq':
/home/winfried/Dropbox/git_normaliz/nmz_opt_lib/Flint_source/flint-2.6.3/fmpq_poly/inlines.c:148: multiple definition of `fmpq_poly_add_fmpq'
/home/winfried/Dropbox/git_develop/local/lib/libeantic.a(EANTIC_fmpq_poly_add_fmpq.o):/home/winfried/Dropbox/git_develop/nmz_opt_lib/E-ANTIC_source/e-antic-1.0.0-rc.12/libeantic/src/../upstream/patched/fmpq_poly_add_fmpq.c:15: first defined here
```

Apart from the problem: wasn't Flint 2.7.0 a major point in this change to e-antic 1.0.0?

saraedum:
As for the first one: I have no idea what's going on. If that line in `missing` says `aclocal-1.16` then it's not something we introduced here. However, there should be no need to run aclocal if this is a release tarball, only running configure should be good enough. It works for me so I have no great idea here to be honest.

The second one: You're trying to build a static libnormaliz I assume. That's not possible without additional linker flags since there are symbol clashes. Something like `--allow-multiple-definition` as a linker flags might do the trick on Linux (but not on macOS.)

FLINT 2.7.0: Yes, e-antic is now compatible with FLINT 2.7.0. But it's not required.

w-bruns:
The first problem with aclocal-1.16 disappears if I install MPFR  from scratch. The information I found in Google indicates trhat it has do with time stamps. But it is unclear to me how.

uvonderohe:
I do not get the error involving `aclocal-1.16`.

Trying to compile with Flint 2.7.1 using
```
FLINT_VERSION="2.7.1"
FLINT_URL="https://www.flintlib.org/flint-${FLINT_VERSION}.tar.gz"
FLINT_SHA256=186e2fd9ab67df8a05b122fb018269b382e4babcdb17353c4be1fe364dca481e
```
in `install_scripts_opt/install_nmz_flint.sh`, I get this error already during the compilation of antic:
```
get_fmpz_mod_poly.c: In function ‘_nf_elem_get_fmpz_mod_poly’:
get_fmpz_mod_poly.c:23:9: error: too few arguments to function ‘fmpz_mod_poly_ze
ro’
         fmpz_mod_poly_zero(pol);
         ^~~~~~~~~~~~~~~~~~
```

What would be the best way to compile Normaliz with Flint 2.7?

w-bruns:
I have looked up

https://github.com/wbhart/flint2/commit/17d26d4c957828f3c98c15ddec85108f841d3438

It seems to me that this fix has become part of Flint2.7.1 and therefore is superfluous in e-antic.

fingolfin:
@saraedum wrote:
> That's not possible without additional linker flags since there are symbol clashes

Could you give a hint what clashes that are? That'd be super helpful (Yes, I could and will try it and probably find out, but it may be some days till I have a chance to do this, and having some idea about what the problem is exactly would allow me to think about possible solutions (as I can do that even when I am taking care of the kids at the playground :-), which I do quite a bit due to the lockdown and the Kindergarten being closed).

w-bruns:
The only clash of symbols is fmpq_poly_add_fmpq (there may be more, but they don't show up). The linker flag  --allow-multiple-definition does indeed help under Linux. 

By the way, we are again dependent on Boost.

saraedum:
> I have looked up
> 
> [wbhart/flint2@17d26d4](https://github.com/wbhart/flint2/commit/17d26d4c957828f3c98c15ddec85108f841d3438)
> 
> It seems to me that this fix has become part of Flint2.7.1 and therefore is superfluous in e-antic.

That's true. It's merged into FLINT. But we do not want to require FLINT 2.7.1 in e-antic yet.

saraedum:
> The first problem with aclocal-1.16 disappears if I install MPFR from scratch. The information I found in Google indicates trhat it has do with time stamps. But it is unclear to me how.

I guess mpfr bootstraps its configure because it sees that something is newer than configure. Internally, mpfr's bootstrap was probably created with aclocal 1.16 which is required in that situation. Just a guess.

saraedum:
> @saraedum wrote:
> 
> > That's not possible without additional linker flags since there are symbol clashes
> 
> Could you give a hint what clashes that are? That'd be super helpful (Yes, I could and will try it and probably find out, but it may be some days till I have a chance to do this, and having some idea about what the problem is exactly would allow me to think about possible solutions (as I can do that even when I am taking care of the kids at the playground :-), which I do quite a bit due to the lockdown and the Kindergarten being closed).

We replace broken implementations in FLINT that cause segfaults for us (and supplement things that are missing from old versions of FLINT.) From our configure.ac:

```
# Check if we need to provide our own nf_init_randtest which was introduced in https://github.com/wbhart/antic/pull/47
AC_REPLACE_FUNCS(nf_init_randtest)
# Unconditionally patch nf_elem_scalar_div_fmpq which has aliasing issues in ANTIC, https://github.com/wbhart/antic/pull/48
AC_LIBOBJ(EANTIC_nf_elem_scalar_div_fmpq)
# Unconditionally patch fmpq_poly_add_fmpq which was fixed in https://github.com/wbhart/flint2/commit/17d26d4c957828f3c98c15ddec85108f841d3438
AC_LIBOBJ(EANTIC_fmpq_poly_add_fmpq)
# Unconditionally patch nf_elem_add_fmpq so it uses our patched fmpq_poly_add_fmpq
AC_LIBOBJ(EANTIC_nf_elem_add_fmpq)
```

saraedum:
> By the way, we are again dependent on Boost.

Yes, we use some boost headers in the `renf_class` and in the `renf_elem_class` interface. Is there a problem with boost in normaliz? (Actually, we pull in too many headers in the interface, let me fix that.)

w-bruns:
Well, Max wrote an own dynamic_bitset class in order to get rid of Boost, but there is no problem in Normaliz to use it.

saraedum:
@w-bruns fyi: I documented a way to turn a conda environment into an executable here https://github.com/conda/conda-pack/pull/159/files#diff-92770a9a8b955dd8aca45c1835511099fce783f050f802a9c39927c08fad53f7R11. The resulting Normaliz "binary" for macOS has 7.8MB btw.

w-bruns:
Thanks! We keep this in mind.

w-bruns:
Could you please also transform

https://github.com/Normaliz/PyNormaliz/tree/new_e-antic ?

I think this is easy for you. But it turned out not so easy for us.

w-bruns:
PS. At present he ci tests of PyNormaliz must fail after the transition to e-antic 1.0.0 since the master  branch of Normaliz is still on 0.1.8.

w-bruns:
In case you haven't started yet: let us try again. I think we understand better now what must be done.

If we should not succeed, we will ask for help again.

saraedum:
Great. I have not started yet.

w-bruns:
We are making progress and can see the light at the end of the tunnel. 

A question: why don't we have constructors for renf_elem_class that depend only on renf* ? Why do we need the std__sharedptr? That makes things more complicated. It means that the cone must have and return this information, for sure in PyNormaliz, but also in other contexts if one wants to make renf_elem_class that fit the existing cone, after the cone has been constructed already.

saraedum:
> A question: why don't we have constructors for renf_elem_class that depend only on renf* ? Why do we need the std__sharedptr? That makes things more complicated. It means that the cone must have and return this information, for sure in PyNormaliz, but also in other contexts if one wants to make renf_elem_class that fit the existing cone, after the cone has been constructed already.

The reason is that each element now explicitly keeps their parent field alive. With just a `renf*` elements have no control over when `renf_clear` is being called.

Why do we want this? It makes it much easier to use `renf_elem_class` in managed languages such as Python. For example, it allows us to enforce that `renf_class` is a unique parent object in a very strong sense (which simplifies interaction with SageMath for example.)

For the Cone example. I think it would be slightly better if cone held a `shared_ptr<const renf_class>` instead of just a `const renf_class*`. I did not want to make that change because it changes the API of Cone. In any case, you can always go back and forth between the two with `renf->shared_from_this()` and `shared.get()`.

w-bruns:
I did not see shared_from_this. But I am adding setting and getting shared_ptr<const renf_class> to the cone.

saraedum:
> I did not see shared_from_this.

It's inherited from the `enable_shared_from_this` base class.

w-bruns:
Everything looks good, but at the end of the test suite ofPyNormaliz I get the message

`A unique factory is leaking memory. 2 objects were created through a C++ unique factory but never released. These objects might be part of a legitimate cache that is (unfortunately) not explicitly released upon program termination as is common in garbage-collocting languages such as Python.`

Where should what be  deleted? Not clear to me. Do we need something like a functiomn "clean-up" at the end of the work? In an interactive shell I never know whether there is a next computation coming up.

w-bruns:
PS. Both Normaliz and PyNormaliz branches "new_e-antic" pushed to GutHub.

saraedum:
> Everything looks good, but at the end of the test suite ofPyNormaliz I get the message
> 
> `A unique factory is leaking memory. 2 objects were created through a C++ unique factory but never released. These objects might be part of a legitimate cache that is (unfortunately) not explicitly released upon program termination as is common in garbage-collocting languages such as Python.`
> 
> Where should what be deleted? Not clear to me. Do we need something like a functiomn "clean-up" at the end of the work? In an interactive shell I never know whether there is a next computation coming up.

In principle, all objects' `__del__` gets called at some point. However, this is not guaranteed to happen, see https://docs.python.org/3/reference/datamodel.html#object.__del__ (but in interactive sessions, this typically happens when the session is closed.)

In my experience, it's likely that something is leaking memory somewhere. For example, if you hold on to one of the shared pointers, you can ask for its [use_count](https://en.cppreference.com/w/cpp/memory/shared_ptr/use_count) before the program terminates. If it's more than one, then there might be a problem. You can invoke the garbage collector explicitly in Python with `import gc; gc.collect()` you might even want to make sure all locals are gone before by doing a very crude `for a in list(locals().values()): del(a)`.

w-bruns:
O.K. I merge into master.

w-bruns:
Please have a look at the function

`static PyObject* _NmzConeIntern_renf(PyObject* kwargs)`

in

https://github.com/Normaliz/PyNormaliz/blob/master/NormalizModule.cpp

Do we really need the  `std::shared_ptr<const renf_class>*` ? Iwanted to make sure that the `std::shared_ptr<const renf_class> ` to which it points survives the end of this function. In former times the `renf_class*` was made by `new`.

saraedum:
The `new` there makes this shared pointer and the underlying number field completely immortal, i.e., leak memory (as of course does any `new` without a `delete`.)

Without the new there, i.e., by just writing 
`auto renf = renf_class::make(poly, var, emb);`, the renf stays alive until it is referenced by anybody holding a shared pointer to it, e.g., as long as any `renf_elem_class` referencing it is around.

The best thing to do here would be for cone's `setRenf` to internally store a `std::shared_ptr<const renf_class>` instead of a `renf_class*`. Then the number field will then survive as long as any cone or element referencing it is alive.

w-bruns:
I will change PyNormaliz following your suggestion.

The cone has the std::shared_ptr already. I hope the following does the job:

```
template <>
void Cone<renf_elem_class>::setRenf(const renf_class* renf) {
    Renf = renf;
    renf_degree = fmpq_poly_degree(renf->renf_t()->nf->pol);
    RenfSharedPtr = renf->shared_from_this();
}
```

saraedum:
Ok. That will work. But why do you store both the `renf_class*` and the shared pointer? It should be good enough to only keep the shared pointer:

```
std::shared_ptr<const renf_class> Renf;

setRenf(const std::shared_ptr<const renf_class>& renf) {
  Renf = renf;
}
```

and if you want to keep the old interface:
```
setRenf(renf_class* renf) {
  Renf = renf->shared_from_this();
}
getRenf() {
  return Renf.get();
}
```

w-bruns:
It was clear to me that the shared_ptr is enough. I kept it for tradition, but it can of course go.

Let me report what we have done, partly changing your versions.

1. We avoid the clash of symbols by patching the two files in the e-antic source in which `fmpq_poly_add_fmpq` appears. we extend the name to `fmpq_poly_add_fmpq_eantic`. The other duplicated symbols are not noticed.
2. We have returned to our ci build shell scripts (with some changes in them).
3. We always install Flint 2.6.3. Compilation with Flint 2.7.1 does not go through, neither with antic 0.2.1, nor with e-antic 1.0.0.
4. Overlinking is completely avoided as suggested by you, with one exception: in the ci build that produces our Mac binary for download. For this purpose we patch configure.ac in the ci build script. So it remains on our private area.
5. We have augmented configure.ac in order by looking for `nauty.h` and `nauty/nauty.h` and anylyzing whether nauty was build with TLS enabled. (Will play a role in 3.9.0.)

We have not yet changed the input functions in Normaliz. The field is still conveyed by the stream. But we will eventually change that.

w-bruns:
**Substantial loss in speed?**

Today we have run Normaliz with e-antic 1.0.0 on our benchmark input. Unfortunately there seems to be a substantial loss in speed. We have compared our A553 input for 3 types of arithmetic:

rat2: purely rational arithmetic in a degree 2 field
scale2: input scaled by nonrational field elements
scale8: similarly for degree 8.

We compare to a run in last June on Winfried's PC with 8 threads. Normaliz tries to measure the speed of the linear algebra by taking the average time of elementary row operations on matrices selected from the input generators. This is measured in "ticks per row", briefly "tpr" in the following. (These times are quite volatile.) As you can see, the ratio new/old becomes smaller when the complexity of the arithmetic rises. This seems to indicate that the loss in speed is caused by an administrative overhead in 1.0.0. We give the times  in the format new/old.

rat tpr 10.01 / 1.70, wall time 21:57 / 4:19
sc2 tpr 11.82 / 4.3, wall time 28: 54 / 12:15
sc8 tpr 19.56 / 13.19 wall time 34:22 / 27:00

For mpz_class arithmetic the new version is actually faster than the old one:

mpz tpr 2.47 / 3.22, wall times 4:04 / 6:54

The official Linix binary 3.8.9 has times very close to those of last summer.

We can provide gperfools measurements if desired.

saraedum:
> We compare to a run in last June on Winfried's PC with 8 threads.

I don't understand. Could you compare to the version just before the e-antic 1.0.0 changes?

saraedum:
>     1. We avoid the clash of symbols by patching the two files in the e-antic source in which `fmpq_poly_add_fmpq` appears. we extend the name to `fmpq_poly_add_fmpq_eantic`. The other duplicated symbols are not noticed.

I don't think that this is enough in general. If you don't replace the `fmpq_poly_add_fmpq` symbol, then antic's `add_fmpq` will still call the broken `fmpq_poly_add_fmpq`.

>     2. We have returned to our ci build shell scripts (with some changes in them).

Ok. You certainly now best.

>     3. We always install Flint 2.6.3. Compilation with Flint 2.7.1 does not go through, neither with antic 0.2.1, nor with e-antic 1.0.0.

I don't understand what you mean exactly. It worked in the CI at some point, didn't it?

w-bruns:
Version 3.8.9 is "just before e-antic 1.0.0 changes". As pointed out, its computation times are essentially the same as those of last  summer.

saraedum:
Knowing which operations exactly are causing trouble would certainly help to improve performance. So if you can point me to the critical lines in your code base and reproducer that has bad performance it's hopefully easy for me to do something about it.

w-bruns:
Two replies to your comments on our changes:

`I don't think that this is enough in general. If you don't replace the fmpq_poly_add_fmpq symbol, then antic's add_fmpq will still call the broken fmpq_poly_add_fmpq.`

You have replaced the broken fmpq_poly_add_fmpq of Flint and we are using the improved version.

`I don't understand what you mean exactly. It worked in the CI at some point, didn't it?`

The CI files have never called Flint 2.7.0/1, only 2.6.3. On Linux, apt gives us a version that is too old, and on Mac we get from Homebrew one that is too new ...

w-bruns:
In regard to the computation times I can already say that the construction and especially the destruction of `renf_elem_class` take a very large amount of time. Precise performance data will follow. Just one detail now: in the "rat" computation with e-antcic 1.0.0 the destruction takes about 44% of the computation time and the construction about 16%. If you subtract this time share from the new computation time, then you get more or less the old one.

w-bruns:
We are attaching 8 files with gperf data. 4 runs in 2 formats each (txt and pdf). old" refers to Normaliz master immediately before the introduction of e-antic 1.0.0, "new" to the runs with e-antic 1.0.0. The two input variants are "rat" and "sc2" as already explained. The input files A553_rat_2.in and A553_scale_2.in are in playground of master.

The conclusion is quite clear: the destruction of renf_elem_class takes an enormous amount of time, and the construction is not much better. This becomes particularly clear if one looks at the "rat" files: ~46% of the time go into the destruction of renf_elem_class, ~22% into the construction. In the "sc2" the share of the destruction and destruction is about 1/2.

Does it make sense to bind a std::shared_ptr to local variables that are created and destroyed at very high frequency? I know that CoCoA uses boost::shared_ptr, but only in the interactive program CoCoA, not in CoCoALib.
[old_rat.pdf](https://github.com/Normaliz/Normaliz/files/5923703/old_rat.pdf)
[old_rat.txt](https://github.com/Normaliz/Normaliz/files/5923704/old_rat.txt)
[old_sc2.pdf](https://github.com/Normaliz/Normaliz/files/5923705/old_sc2.pdf)
[old_sc2.txt](https://github.com/Normaliz/Normaliz/files/5923706/old_sc2.txt)
[new_rat.pdf](https://github.com/Normaliz/Normaliz/files/5923708/new_rat.pdf)
[new_rat.txt](https://github.com/Normaliz/Normaliz/files/5923709/new_rat.txt)
[new_sc2.pdf](https://github.com/Normaliz/Normaliz/files/5923710/new_sc2.pdf)
[new_sc2.txt](https://github.com/Normaliz/Normaliz/files/5923711/new_sc2.txt)

saraedum:
Sorry if that's a stupid question. But can you give me the exact command lines that I need to run these examples? (I just never use normaliz myself.)

uvonderohe:
If I may answer that question: The command should be a plain `./normaliz playground/A553_rat_2` and analogously for the other example. You can enable verbose mode by adding the option `-c`.

saraedum:
I had a look at the first part, the `tpr`. With a few minor changes in Normaliz I get this down from 12-15 to about 4 for A553_rat_2.in. With some changes in e-antic even to 2.5. So, while it's true that allocations of `renf_elem_class` are more expensive now than they used to be, that only accounts for a small part of the problem, and can be mitigated by changing two expressions in the row echelon algorithm. The main issue seems to be that working with rational elements is much slower in some cases. This happens because we removed a lot of the special casing around this in e-antic (before, a `renf_elem_class` was essentially a `variant<mpq_class, number field element>`.)

> rat2: purely rational arithmetic in a degree 2 field

With "purely" you mean that the computation would work equally with `mpq_class` instead? If so, then this is of course the worst case for the kind of changes described above. Of course we should make this faster than it is now, but I am not sure how many hacks we want to add to e-antic to special case for rational elements. @videlec

saraedum:
An example where it's going to be difficult to improve the situation: Of the `tpr`, 22% are now spent in `Matrix::select_submatrix`. All this does is copy `renf_elem_class` from one matrix to another. Before, this, we only had to copy a `fmpq_t`. Now this copies the degree 2 polynomial describing the rational and the ball giving an approximate embedding. The two operations are where 98% of the time are spent in the `operator=` here. Only 2% are caused by number field things such as the shared pointer overhead.

w-bruns:
What you have written, is very promising. In case you want to make a PR in the near future, please direct it to `new_e-amtic_in_master`.

Some remarks from the Normaliz side:

1) `rat` is a Spielerei. Normaliz does not really need a `Cone<mpq_class>` which is (or was) imitated by `rat`.

2) The` tpr` help to select algorithmic variants. The main point is the choice between the rank test and the comparison test for codim 2 faces. The first is a rank computation, the second compares incidence vectors. For the rank test we use prefabricated matrices that belong to the `Full_Cone` and must be constructed only once. The vectors spanning the critical face are copied to such a matrix.

 The measurement for `tpr` does not use the prefabricated matrices. Perhaps we should do it in order to get a more precise measurement.

3) Can one do something for the scalar product? The basic operation is `sc_prod += a[i]*b[i]`. I am quite sure that one has to construct a renf_elem_class for the product on the right hand side. Would` p = a[i]   p *= b[i]  sc_prod += p` with a prefabricated `p` be any better?

 Some time ago I raised the question (in a mail to Vincent or perhaps even in an issue on GitHub) whether it could make sense to have a special function for scalar products. In fact, the reduction modulo the minimal polynomial needs to be done only once for the final result, not for every `a[i]*b[i]`. I am not sure what effect this could have on the `arb`.

w-bruns:
I have pushed `new_enatic_in_master` to GitHub with the change in `ramk_time()` that makes it use prefabricated matrices. Let's see what effect this has.

A correction to what I wrote above: the times are not used in the decision "rank test" vs. "comparison test" in general, but for "small pyramid vs. large pyramid" and in large pyramids for "rank test" vs. "comparison test".

w-bruns:
I have experimented with `A553_scale_2` which is much more important than rat.

I guess that Julian changed the Normaliz code in order to reduce the construction of `renf_alem_class` for local intermediate results. I have done this in `reduce_row(...)`:

```
        help1 = help;
        help1 *= elem[row][j];
        elem[i][j] -= help1;
```

using `help` and `help1` as "registers". This brings the `tpr` down from ~11 to ~6. However, the computation increased from ~28 min to ~31 min. This negative effect was caused by a shift from "large pyramids" to "small pyramids". In order to avoid this influence I have simply set the `tpr` to 11. Then we are essentially back at the old computation time of ~28 min.

As a next step I have applied the change above also in `solve_destructive_inner(...)`: 

 ```
          for (int k = i - 1; k >= 0; --k) {
                fact = elem[k][i];
                for (size_t j = i; j < nc; ++j){
                    help = elem[i][j];
                    help *= fact; 
                    elem[k][j] -= help;
```

This change had a significant effect: the computation time went down to ~21 min. So we are halfway back to the time with e-antic 0.1.8.

The same trick in `v_scalar_product(...)` did not work: it increases computation time. But I will try again.

saraedum:
Thanks for looking into this also. I had also found that making the tpr faster made everything slower. I had assumed that I had broken something else in the process.

I was organizing a conference this week so I could not look into this yet. I'll try to make some progress here next week.

w-bruns:
I have done some changes that shifts pyramids from small to large:  `ticks_rank_per_row *= 3` and earlier use of `small_vs_large()`. This gives a time < 19 min. Tomorrow I will produce a gperf log.

The changes have been pushed to GitGub, branch `new_e-antic_in_master.`

There is still some room for using prefabricated matrices and vectors, but this requires some extensive changes in the code.

w-bruns:
I have pushed the branch again. The change in `v_scalar_product` does not seem to have much influence. It is probably better to take it back.

gperf data attached. The destructor or `refrn_elem_class` still takes a considerable amount of time. As already said, the number of constructions/destructions can be reduced somewhat.

[sc2_Feb_21.pdf](https://github.com/Normaliz/Normaliz/files/6018107/sc2_Feb_21.pdf)
[sc2_Feb_21.txt](https://github.com/Normaliz/Normaliz/files/6018108/sc2_Feb_21.txt)

w-bruns:
The version in new_e-antic pushed some minutes ago brings the time down to ~ 17:50 minutes (on my PC, 8 threads).
The last step one can do on the Normaliz side is the recycling of FACET_DATA. This has been an issue for some time already.

w-bruns:
Unfortunately the recycling of FACRETDATA slows down the computation ... I don't see a real improvement in Normaliz anymore.

w-bruns:
Matthias Köppe asked whether it could make sense to make the type of pointer of a `renf_elem_class` a template.

I am not sure what complications this could cause in the implementation, but it could be a solution for transient data.

w-bruns:
We have upgraded Normaliz/master to Flint 2.7.1 and e-antic 0.1.9 and transferred the improvements in Normaliz to master. As expected they also speed up the "old" e-antic. For example, A663_scale_2 is now < 10 minutes (in the same setup in which the times so far have been measured).

PR #370, opened 2021-03-03
speed up A553_scale_2

saraedum:
Sampling a run of A553_scale_2, I found several things that I could improve. Most importantly, I found that there are lots of calls `a*=b` where one of the operands is zero and calls `a-=b` where the right hand side is zero.

We are going to make such calls much faster in e-antic itself but it is beneficial to not perform any arithmetic in such cases. Note that the previous version of e-antic used a faster path of arithmetic here because it immediately identified such operands as `fmpq` rationals.

For me this brings down the runtime of A553_scale_2 from

```
real    27m7.131s
user    211m21.847s
sys     0m12.758s
```

to

```
real    15m10.432s
user    116m12.130s
sys     0m9.882s
```

### DIFF
--- a/source/libnormaliz/full_cone.cpp
+++ b/source/libnormaliz/full_cone.cpp
@@ -625,7 +625,7 @@ void Full_Cone<Integer>::add_hyperplane(const size_t& new_generator,
 
     // check_facet(NewFacet, new_generator);
 
-    NewHyps.push_back(NewFacet);
+    NewHyps.emplace_back(std::move(NewFacet));
 }
 
 //---------------------------------------------------------------------------
--- a/source/libnormaliz/full_cone.cpp
+++ b/source/libnormaliz/full_cone.cpp
@@ -2222,7 +2222,7 @@ void Full_Cone<Integer>::select_supphyps_from(list<FACETDATA<Integer>>& NewFacet
 template <typename Integer>
 void Full_Cone<Integer>::match_neg_hyp_with_pos_hyps(const FACETDATA<Integer>& Neg,
                                                      size_t new_generator,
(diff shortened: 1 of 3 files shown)


saraedum:
cc @videlec

w-bruns:
Sounds good! I  have looked at the first finished and failed test. There seems to be a problem in the linear algebra. But let us wait for the other tests.

By the way, he branch simplified_reconciliation uses e-antic 0.1.9 and Flit 2.7.1.

saraedum:
I had written `reduce_row` without thinking about floor division that happens for integers. It should be fixed now by completely special-casing e-antic there.

w-bruns:
I have just run A553_scale_2 on my PC. Just 9:07 min, which is faster than 0.1.9 in the branch simplified_reconciliation. Of course, the latter will very likely also profit from the changes in this pull request, so that one should compare again when they have been transferred there.

w-bruns:
e-antic 1.0.0 should now be made compatible with Flint 2.7.1. This requires the exchange of antic release 0.2.1 to tagged version 0.2.3, but also slight changes in e-antic himself. 

Then we will upgrade the next Normaliz release to e-antic 1.0.0.

w-bruns:
All developments of Nornaliz, including the transfer to e-antic 1.0.0-rc-13 have been merged into the branch new_e-antic_in_master. If there are no objections from the e-antic side, we can make it version 3.9.0 in the next two weeks.

saraedum:
We have one more performance fix that changes the API slightly. I'll try to make the change to e-antic during the next week and create a PR for Normaliz. After that we'll release e-antic 1.0.0 finally.

w-bruns:
Very good. Please direct the PR at new_e-antic_in_master.

PR #389, opened 2022-04-01
CI: use ccache, possibly fix Codecov issues

fingolfin:
Using ccache should speed up CI quite a bit


fingolfin:
Any speedups will only be visible after repeated runs. So, to test if it works, it first needs to be merged; then a full CI run for `master` must complete. Then one can open a new PR (can be trivial), and watch its timings (and also check the ccache logs, which include statistics for how effective it was)

PR #390, opened 2022-04-01
CI: add one more ccache invocation

fingolfin:


fingolfin:
This is not yet as effective as it could be due to cache conflicts. I'll tweak it some more...

PR #391, opened 2022-04-02
Update some URLs from http to https

fingolfin:


fingolfin:
@w-bruns this shows that `ccache` works quite well now. E.g. in https://github.com/Normaliz/Normaliz/runs/5785600401?check_suite_focus=true it took 17 minutes 26 seconds to build the prerequisites. Now it is down to 3 minutes 12 seconds in https://github.com/Normaliz/Normaliz/runs/5798951624?check_suite_focus=true  (of course these numbers fluctuate as always, but it is clear we've reduced it substantially)

w-bruns:
I really appreciate this development. 

We could also think of replacing our own installation of packages like Flint by apt/brew. Flint takes a very long time to compile. This was a little difficult when e-antic was upgraded to 1.0.0, but I think the situation has calmed down now.

I is unclear to me whether we can make CodeCov  work again.

fingolfin:
Unfortunately Ubuntu's flint is stuck at 2.6.3 even in version 21.10, see https://packages.ubuntu.com/impish/libflint-dev -- though I think it'll go to 2.9.4 in 22.04, which should be out "soon"...

PR #393, opened 2022-06-02
Compiler warnings and loss of precision in integer assignments

baburton:
Hi! I'm on a long-term mission to reduce the number of compiler warnings on macOS (Xcode/clang) when building regina, so that important messages do not get lost in the noise. At the moment there are hundreds of warnings coming from normaliz, and so I've gone through and audited the normaliz code to quieten things down.

The vast majority of these warnings come from implicitly casting a larger integer type to a smaller type (and the majority of *these* come from implicitly casting size_t to a key_t). In most of these cases I've added a static_cast, which has two effects: the warnings are silenced, but also this is easily searchable if you ever plan to go through and make your use of integer types more consistent across normaliz. (I'm currently doing the same with regina, but it's a slow and ongoing process.) Importantly, these static_casts are not changing behaviour at all - it just makes them explicit instead of implicit, so the loss of precision is easier to spot. (See commits 759085d, 8954583 and 6466c0e).

I do appreciate that to a large degree this is theoretical - if you are trying to compute a Hilbert basis (for example) on a cone whose dimension is larger than INT_MAX then you have bigger problems to deal with. Having said that though, there are systems on which INT_MAX is 32767, so maybe on some systems this does have some practical relevance. However, again, just the benefit of silencing hundreds of compiler warnings is enormously helpful (and my main motivation here).

There are some places where I have upgraded types from int to something larger, mostly where variables being changed are sufficiently local that it's easy to check that nothing should break. (See commits de3d014 and c589551.) For OMP thread counts I've downgraded variables from size_t/long to int, since OpenMP itself uses int for this purpose (see commit 1a339c1). Some of the upgrades are from int to ssize_t, which is a type that is not in the C++ standard but which you already have access to in normaliz since you are already including sys/types.h (see commit c589551) - this is mostly used for cases where you loop down to zero, since ssize_t is allowed to take the value -1.

The only commit I have not mentioned was changing a (const bool&) to just a plain bool-by-value inside a range-based for loop, since this is simpler and probably cheaper also (commit 08c495e).

Anyway, that's all! Please do let me know if you have any questions about any of this. - Ben.

### DIFF
--- a/source/libnormaliz/HilbertSeries.cpp
+++ b/source/libnormaliz/HilbertSeries.cpp
@@ -145,7 +145,7 @@ void poly_div(vector<Integer>& q, vector<Integer>& r, const vector<Integer>& a, 
     r = a;
     remove_zeros(r);
     size_t b_size = b.size();
-    int degdiff = r.size() - b_size;  // degree differenz
+    size_t degdiff = r.size() - b_size;  // degree differenz
     if (r.size() < b_size) {
         q = vector<Integer>();
     }
--- a/source/libnormaliz/HilbertSeries.cpp
+++ b/source/libnormaliz/HilbertSeries.cpp
@@ -690,7 +690,7 @@ void HilbertSeries::computeHilbertQuasiPolynomial() const {
     for (j = 0; j < reduced_period; ++j) {
         INTERRUPT_COMPUTATION_BY_EXCEPTION
 
-        quasi_poly[j] = compute_polynomial(quasi_poly[j], dim);
+        quasi_poly[j] = compute_polynomial(quasi_poly[j], static_cast<int>(dim));
     }
 
     // substitute t by t/period:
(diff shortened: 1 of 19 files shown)


w-bruns:
Thanks for making your concerns so precise! I have been feeling uneasy with the integer types for a long time. I think the distinction of `key_t` and `size_t` comes from the time when 32 bit systems were widely in use and memory could be a problem. (On MS Windows `long` and `long long` doe still differ.) 

I hope that I can devote myself to this cleaning up in the second half of this year --- if there is anything left after your efforts.

w-bruns:
There is a significant extension now in the branch `with_markov`, containing the computation of Markov and Gröbner bases of lattice ideals plus Hilbert series of the quotient. It would be great if you could have a look at it. Some of the basic stuff was written by uvonderohe. Your changes to `master` have been transferred to `with_markov`.

PR #396, opened 2022-07-06
Don't inline getVersion()

fingolfin:
An inline version of getVersion() is next to useless, one can simply
access NMZ_RELEASE directly.

A non-inline version however can be used to compare the runtime version
of libnormaliz to the compile time version. That can be beneficial to
catch certain kinds of bugs.

### DIFF
--- a/source/libnormaliz/general.cpp
+++ b/source/libnormaliz/general.cpp
@@ -138,4 +138,9 @@ void MeasureTime(bool verbose, const std::string& step) {
 }
 #endif
 
+unsigned int getVersion()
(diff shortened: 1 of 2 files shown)


w-bruns:
Two checks on Mac OS have failed because of problems with the installation of e-anric. They have to do with a replacement function of e-antic for a Flint function. It could be that it is no longer needed.

Why this happens now (and has not happened in the last checks of master on June 2) is unclear to me. I merge this PR. The e-antic problem must be analyzed later.

PR #402, opened 2022-12-09
Fix typos found by codespell

fingolfin:


w-bruns:
It seems that at least one comment has been de-cpmmented. I am hesitant to merge this PR .

fingolfin:
This patch was 99% created by an automatic tool. But it reported a few places where it noticed a spelling mistake but could not fix it automatically (e.g. `ther` could be a typo for `there`, `their` or `the`), which had to fix manually. In this case I assume I accidentally slipped and hit 'enter' instead of backspace and didn't notice.

I don't see how anything worse than this (easily diagnosed and fixed) issue could crop up.

w-bruns:
Thanks!

PR #408, opened 2023-05-25
Don't use -static for macOS builds

fingolfin:


fingolfin:
Ah, my bad, I also need to make sure our scripts find that LLVM. I'll update the PR shortly.

fingolfin:
OK it now picks up clang 15 (before it picked up Apple clang 14). Still errors. Though I understand why, I also need to adjust the include and library dirs.



However, I also noticed in both logs the message
> `warning: argument unused during compilation: '-all-static'`

That suggest that `-all-static` is not supported on any of them. I looked at multiple older versions of clang, down to version 10, and none of them supported -all-static, at least according to their manuals. Where does that come from then?!

Anyway this suggests to me that perhaps this PR is going in the wrong direction. Instead (or in addition?) we need to replace `--all-static` but by what? Maybe `-static`?

I also now wonder if any recent macOS release binaries for normaliz actually work in practice?

w-bruns:
It could be that there is an additional -static used. Perhaps all-static was simply ignored so far, but is now an error. I will check the install files.

w-bruns:
I have replaced `-all-static` by `-static` in the last push to master. <this seems to be o.k. Nevertheless configure crashes. It is unclear to me what the problem is. Please have a look.

fingolfin:
The error causing `configure` to fail is this:
```
configure:4591: clang  -I/Users/runner/work/Normaliz/Normaliz/local/include  -I /usr/local/opt/llvm/include -I/usr/local/include -L/Users/runner/work/Normaliz/Normaliz/local/lib  -L/usr/local/opt/llvm/lib -Wl,-rpath,/usr/local/opt/llvm/lib -L/usr/local/lib -static conftest.c  >&5
ld: library not found for -lcrt0.o
clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
```


And that's not unexpected, see https://stackoverflow.com/questions/3801011/ -- basically `-static` never was useful on macOS.

The usual solution is to compile all required libraries oneself (I think we do that for this very reason) as static libraries. Then either make sure no shared library variants of them are around (e.g. build them all with `--enable-static --disable-dynamic`), or else link explicitly against the static variants (so instead of doing `-L/some/path/to/lib -lfoo` do `/some/path/to/lib/libfoo.a`.

That's how I've done it for other projects for a long time until a couple years ago. This is also what random web results seem to suggest, e.g. https://jonwillia.ms/2018/02/02/static-linking

Looking at the various scripts in the Normaliz repository, I think this is already being done. E.g. I see this:
```
        # Remove shared libraries and libtool *.la files to force static linking
        # ls -laR ${OPTLIBDIR}
        rm -f ${OPTLIBDIR}/*.dylib*
        rm -f ${OPTLIBDIR}/*.so*
        rm -f ${OPTLIBDIR}/*.la
        # ls -laR ${OPTLIBDIR}
        if [[ $OSTYPE == darwin* ]]; then
            BREWDIR=$(brew --prefix)
            rm -f ${BREWDIR}/lib/*gmp*.dylib*
            rm -f ${BREWDIR}/lib/*mpfr*.dylib*
            rm -f ${BREWDIR}/lib/*flint*.dylib*
        fi
```

So here's my theory: the `-all-static` flag was not doing anything. So the correct fix is to remove it, and that's it -- replacing it by `-static` is wrong. So the new fix then is to remove `-static`... I've just changed this PR to do just that. Let's see how it works out...

w-bruns:
I have looked again at the install scripts, and, honestly, I am not sure that under MacOS they do exactly what is promised in the manual. I am quite sure that under Linux everything works as stated in the manual.

w-bruns:
I think the reasonable strategy for MacOS is the following (and I think it is intended in the install scripts, but perhaps not clear in the manual): 

1) For the binary distribution we make a (almost) static binary as done already by the workflow.

2) If someone builds Normaliz on the own hardware, then a shared binary is good enough (and most likely conforms with the Apple ideology).

3) For 2) it should be possible to use the Homebrew compiler as well as the Apple compiler. At present he Apple compiler is used in the build based on our own install scripts, and it works.  I will analyze this build and also try to use the Homebrew compiler there.

w-bruns:
I have changed the install scripts build under MacPOS to the Homebrew compiler. Moreover,` NMZ_SHARED=yes `is set for it.  Works correctly. It remains to check and to update the manual if necessary.

fingolfin:
Very good

PR #415, opened 2024-03-02
Fix typos

d-torrance:
This fixes a couple of typos that were flagged by Debian's `lintian` tool while working on the 3.10.2 package.

### DIFF
--- a/source/libnormaliz/output.cpp
+++ b/source/libnormaliz/output.cpp
@@ -1468,7 +1468,7 @@ void Output<Integer>::write_files() {
         }
         if (homogeneous && Result->isComputed(ConeProperty::IsSerreR1)) {
             if (Result->isIntegrallyClosed()) {
-                out << "original monoid satifies Serre cindition R1" << endl;
+                out << "original monoid satisfies Serre cindition R1" << endl;
             }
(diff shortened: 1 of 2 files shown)


source/libnormaliz/output.cpp
```diff
@@ -1468,7 +1468,7 @@ void Output<Integer>::write_files() {
         }
         if (homogeneous && Result->isComputed(ConeProperty::IsSerreR1)) {
             if (Result->isIntegrallyClosed()) {
-                out << "original monoid satifies Serre cindition R1" << endl;
+                out << "original monoid satisfies Serre cindition R1" << endl;
             }
             else {
                 out << "original monoid vioaltes Serre cindition R1" << endl;
```
jamesjer, line 1474: Two more misspelled words appear in the context of this patch: "vioaltes" on line 1474 should be "violates".  The same misspelling occurs in test/test-grb/latt_ideal_grad.ref.  Also, all of the "cindition" instances visible here should be "condition".

d-torrance: Thanks for noticing these!  I've just pushed two commits fixing these additional typos.

PR #418, opened 2024-05-08
Fix macOS CI by installing more dependencies explicitl

fingolfin:
It seems GitHub recently changed which packages they install by default in their macOS runners.

This may need some further tweaks, so please let's wait till it completes before merging, to make sure it is ready (if not I'll adjust it further)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -215,9 +215,9 @@ jobs:
       NO_COVERAGE: "1"
     steps:
       - uses: actions/checkout@v3
-      - name: "Setup compiler and build automake"
+      - name: "Install homebrew packages"
         run: |
-          brew reinstall llvm automake boost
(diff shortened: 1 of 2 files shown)


fingolfin:
OK, second problem: the new Homebrew is no longer install in `/usr/local` but rather in `/opt/homebrew` (the old location is not possible anymore in recent macOS versions).

So we probably need to add the right prefix derived from `$(brew --prefix)` to some CPPFLAGS, LDFLAGS etc., and/or teach some configure scripts to automatically find it there.

w-bruns:
The tests that fail, depend on computations that are not completely deterministic. I will check whether the results are in the a priori error bounds. 

The build with "static" in its name fails because kibomp.dylib can't be found.

w-bruns:
The results are within the error bounds. I think  I must disable the comparison of fixed precision numbers.

PR #434, opened 2025-01-28
adjust FLINT download url

thchr:
It seems the FLINT download url has changed, with a "/download" now added to the address. Noticed when I was trying to compile a fresh version today.


thchr:
Was there a problem with this change?

w-bruns:
Thanks, no problems at all. Good that you found this change.

thchr:
Ah: I think you maybe pressed "close" instead of "merge" by accident?

w-bruns (approved):
Everything o.k.

w-bruns:
Sorry, I thought I had merged it. I think I got lost in the forest of safety checks.

fingolfin:
I've merged it now. Thank you!

PR #453, opened 2026-06-13
Reduce the number of included CoCoALib headers

jamesjer:
Fixes https://github.com/Normaliz/Normaliz/issues/452.  This PR includes only those CoCoALib headers that are really needed, instead of pulling them all in via `CoCoA/library.H`.  I have verified that Singular, polymake, and the GAP NormalizInterface package all build successfully after building normaliz with this change.


w-bruns (commented):
O.K.

w-bruns (approved):

